PR #1, opened 2020-01-30
Update .gitignore

zeinalii:


.gitignore
```diff
@@ -9,6 +9,7 @@
 
 venv/**
 settings.json
+db.sqlite3
```
zeinalii: this should be in the git ignore

PR #31, opened 2020-09-26
Renamed event column in problem set to Type

rachellegelden:
image


rachellegelden:
Closing. Updating where renaming of "Event" happens to a common location so that "Event" is replaced with "Type" in Create New Problem as well

keyvankhademi (approved):

PR #33, opened 2020-09-29
Have a Consent Form while Signing Up

carson-ricca:
- Added consent form on signup.
- Works when clicking the checkbox or the url.
     - Opens the modal and the user can cancel or consent to the form.
- If the user clicks the cancel button the checkbox will remain unchecked, if the click 'I Consent' the checkbox will be checked and they can submit the form.


accounts/static/accounts/js/register.js
```diff
@@ -0,0 +1,23 @@
+$(function () {
+    $("input[open_modal=\"true\"]").on("change", function () {
```
keyvankhademi, line 2: You can have $("input[open_modal=\"true\"]") as variable at the top so it doesn't search for it every time. Make the performance better.

templates/terms_and_conditions_snippet.html
```diff
@@ -0,0 +1,76 @@
+<h4>INFORMED CONSENT</h4>
```
keyvankhademi, line 1: We made this so we could use it in our terms and conditions page. Please include this file there too.

keyvankhademi (approved):

PR #47, opened 2020-10-06
Adds Indents to Parson Problems

carson-ricca:
- Automatically adds indents to parsons problems code after an open bracket.
- Reformatted code in the files I touched.
- Added CSS to decrease the indent size in parsons problems for visual appeal.
- Fixed the settings.py to allow account creation on local machine.


course/templates/parsons_question.html
```diff
@@ -1,47 +1,103 @@
...
-            $('#source_code').val(code);
-        }
-    </script>
-    <div class="my-3">
-        <form method="post" enctype="multipart/form-data">
-            {% csrf_token %}
+    function calculate_source_code() {
+        let parsons_lines = $(".right-container pre.code_lines");
+        let code = "";
+        parsons_lines.each(function (i, e) {
```
keyvankhademi, line 43: $("source_code").val(parson_lines.map(function (){return $(this).text()}).get().join("\n"))

I tried to write this in a compact format, but it looks so ugly. I'm not sure which one is better :/

course/templates/parsons_question.html
```diff
@@ -1,47 +1,103 @@
...
+        $("#source_code").val(code);
+    }
 
-            <input type="hidden" id="source_code" name="code" class="form-control">
+    function determine_indents() {
+        let parsons_lines = $(".right-container pre.code_lines");
+        let count = 0;
+        parsons_lines.each(function (i, e) {
+            let line_value = $(e);
+            let line = line_value.text().trim();
```
keyvankhademi, line 56: The lines are not trimmed in the backend, are they?
I think we have to trim it in the back-end too. and keep this just to be safe.

carson-ricca:
I don’t think they are trimmed in the backend

course/templates/parsons_question.html
```diff
@@ -1,47 +1,103 @@
...
+            }
+        });
+    }
 
-            <div class="form-group my-3">
-                <button type="submit" class="btn btn-success"> Submit</button>
-            </div>
-        </form>
-    </div>
+    function remove_indents() {
```
keyvankhademi, line 72: can you please rename this to remove_left_container_indents.

course/templates/parsons_question.html
keyvankhademi: now I see why you needed this.

course/templates/parsons_question.html
carson-ricca: Ahhh I'm not too sure what's better

course/templates/parsons_question.html
```diff
@@ -1,47 +1,103 @@
...
+        let parsons_lines = $(".left-container pre.code_lines");
+        parsons_lines.each(function (i, e) {
+            let line_value = $(e);
+            let line = line_value.text().trim();
+            line_value.text(line);
+        });
+    }
+
+    function indent(str, numOfIndents, opt_spacesPerIndent) {
+        str = str.replace(/^(?=.)/gm, new Array(numOfIndents + 1).join("\t"));
```
keyvankhademi, line 82: I have no idea what this function does and can't figure it out from the code.
maybe add some comments.

course/templates/parsons_question.html
carson-ricca: Yes

course/templates/parsons_question.html
carson-ricca: Okay sounds good

course/static/course/css/parsons.css
```diff
@@ -0,0 +1,9 @@
+.code_lines {
```
keyvankhademi, line 1: these two are identical. A better way to do this is to make it one with a more reasonable name like "ts-2".
and if the classes are not used in any other way replace it with ts-2 otherwise add a ts-2 class to the elements you need.

course/templates/code_submission_detail.html
```diff
@@ -49,7 +50,7 @@ <h1>Submission Detail</h1>
 
     <h2> Code </h2>
     <pre>
-        <code class="language-java">{{ submission.answer }}</code>
+        <code class="language-java" id="parsons_submission_detail">{{ submission.answer }}</code>
```
keyvankhademi, line 53: here you can just replace this with ts-2

course/templates/parsons_question.html
```diff
@@ -1,47 +1,103 @@
...
-                        <div class="container-object">{{ line }}</div>
-                    {% endfor %}
+<div class="container">
+    <div class="row">
+        <div class="col-md-6">
+            <h5>Lines</h5>
+            <div class="left-container">
+                {% for line in uqj.get_lines %}
+                <div class="container-object">
+                    <pre class="code_lines my-0">{{ line }}</pre>
```
keyvankhademi, line 19: same as here

course/templates/parsons_question.html
keyvankhademi: leave it for now, its more readable.

keyvankhademi (approved):

PR #62, opened 2020-10-21
Course view

Opeyem1a:
- Populating the grade field in event snippet of course view
- Also adding create new event form


Opeyem1a:
Note: Doesn't implement the "Import Event" button.

canvas/forms/forms.py
```diff
@@ -0,0 +1,37 @@
+from django import forms
+from django.forms import TextInput, CheckboxInput, DateTimeInput, Select
+from canvas.models import CanvasCourse
+from datetime import datetime
+
+
+class CreateEventForm(forms.Form):
```
keyvankhademi, line 7: Try using forms.ModelForm for forms that are creating a model.

canvas/views/views.py
```diff
@@ -78,3 +81,25 @@ def events_options_view(request):
...
         'events': course.events.all(),
     })
+
+
+def create_event_view(request):
+    if request.method == 'POST':
+        form = CreateEventForm(request.POST)
+        if form.is_valid():
+            try:
+                create_event(name=form.cleaned_data['name'], course=form.cleaned_data['course'],
```
keyvankhademi, line 91: Using a ModelForm, here you can just say form.save(), and there no need for create_event function.

canvas/views/views.py
```diff
@@ -78,3 +81,25 @@ def events_options_view(request):
...
+                             count_for_tokens=form.cleaned_data['count_for_tokens'],
+                             start_date=form.cleaned_data['start_datetime'],
+                             end_date=form.cleaned_data['end_datetime'])
+                messages.add_message(request, messages.SUCCESS, 'Event created successfully')
+            except EventCreateException:
+                messages.add_message(request, messages.ERROR, 'Error in event creation.')
+
+            return HttpResponseRedirect('.')
```
keyvankhademi, line 99: I'm not sure what '.' does in HttpResponseRedirect. I don't think you need to redirect here you have to stay on the same page to display errors in the form.

course/templatetags/event_grade.py
```diff
@@ -0,0 +1,20 @@
+from django import template
```
keyvankhademi, line 1: why is this template tag is in the course app if you are using it in the canvas app?

canvas/utils/utils.py
```diff
@@ -1,5 +1,30 @@
...
+def create_event(name=None, course=None, count_for_tokens=None, start_date=None, end_date=None):
+    try:
+        from canvas.models import Event
+        new_event = Event(name=name, course=course,
+                          count_for_tokens=count_for_tokens,
+                          start_date=start_date,
+                          end_date=end_date)
+        new_event.save()
+    except Exception as e:
+        print(e)
```
keyvankhademi, line 19: do not print. Use a logger if you want to log something so we can see it in the logs.

canvas/urls.py
```diff
@@ -1,13 +1,14 @@
...
-from canvas.views.views import course_list_view, course_view, event_problem_set, events_options_view
+from canvas.views.views import course_list_view, course_view, event_problem_set, events_options_view, create_event_view
 
 urlpatterns = [
     path('<int:pk>', course_view, name='course'),
     path('<int:pk>/register', register_course_view, name='course_register'),
     path('events-options', events_options_view, name='course_events_options'),
     path('event/<int:event_id>/problem-set', event_problem_set, name='event_problem_set'),
     path('', course_list_view, name='course_list'),
+    path('create-event', create_event_view, name='create_event')
```
keyvankhademi, line 12: create-event is per course if I'm not mistaken. When you click on create event in a course it should create that event for that specific course. I saw that you have course as option in the form. We also want to restrict event creation to the course owners.

canvas/forms/forms.py
```diff
@@ -0,0 +1,37 @@
...
+            attrs={'class': 'form-control'}
+        ),
+    )
+    count_for_tokens = forms.BooleanField(
+        label="Does this event count for tokens?",
+        widget=CheckboxInput(attrs={'class': 'form-control'}),
+        required=False
+    )
+
+    start_datetime = forms.DateTimeField(
```
keyvankhademi, line 27: I would recommend using a good datetime picker widget. 
You can try this one https://www.jqueryscript.net/time-clock/Lightweight-jQuery-Date-Input-Picker.html

keyvankhademi: Actually that one might not have time in it.
https://xdsoft.net/jqplugins/datetimepicker/
try this one in that case.

course/templatetags/event_grade.py
```diff
@@ -0,0 +1,20 @@
...
+from course.models.models import UserQuestionJunction
+
+register = template.Library()
+
+
+@register.filter
+def total_event_grade(event, user):
```
keyvankhademi, line 8: This is really inefficient. It will slow than the website as the number of uqjs are a lot.
I recommend using aggregate functions you learned in 304.
also you start the filter on user.question_junctions rather than starting from the whole table.

keyvankhademi:
also make sure the pipeline is passing.

canvas/templates/canvas/event_create.html
```diff
@@ -0,0 +1,47 @@
...
+                {{ field }}
+                {% if field.help_text %}
+                    <small class="form-text text-muted">{{ field.help_text }}</small>
+                {% endif %}
+            </div>
+        {% endfor %}
+        <input class="btn btn-success mt-md-5" type="submit" value="Submit">
+    </form>
+    <!-- XDSoft DateTimePicker -->
+    <link rel="stylesheet"
```
keyvankhademi, line 41: These links should be in the script block. Also, you need to download the files, put it in the static folder and use {% static %} tag to load them in the script block.

canvas/templatetags/event_grade.py
```diff
@@ -0,0 +1,24 @@
+from functools import reduce
+from django import template
+from django.db.models import Sum, F, Count
+from course.utils.utils import get_token_value
+
+register = template.Library()
+
+
+@register.filter
+def total_event_grade(event, user):
```
keyvankhademi, line 10: It's better not to have this much logic in the template tags. What I would recommend is to create a function in utils to calculate the total event grade and call it here in the template tag.

canvas/utils/utils.py
```diff
@@ -1,3 +1,5 @@
+
```
keyvankhademi, line 1: Can you please change this back?

canvas/views/views.py
```diff
@@ -78,3 +78,26 @@ def events_options_view(request):
     return render(request, 'canvas/course_event_options.html', {
         'events': course.events.all(),
     })
+
+
+def create_event_view(request, pk):
+    course = get_object_or_404(CanvasCourse, pk=pk)
+    course.is_instructor(request.user)
```
keyvankhademi, line 85: I don't understand why this line is here. I think it doesn't do anything. If so please remove this line.

canvas/views/views.py
```diff
@@ -78,3 +78,26 @@ def events_options_view(request):
...
+    if request.method == 'POST':
+        if course.is_instructor(request.user):
+            form = CreateEventForm(request.POST)
+            if form.is_valid():
+                new_event = form.save(commit=False)
+                new_event.course = course
+                new_event.save()
+                messages.add_message(request, messages.SUCCESS, 'Event created successfully.')
+        else:
+            messages.add_message(request, messages.ERROR,
```
keyvankhademi, line 95: permissions should be handled with permission classes or functions. You should not manually do this.

canvas/templates/canvas/event_create.html
```diff
@@ -1,7 +1,12 @@
 {% extends 'base.html' %}
 {% load static %}
 
+{% block style %}
+    <link href="{% static 'css/jquery.datetimepicker.min.css' %}" rel="stylesheet">
+{% endblock %}
+
 {% block script %}
+    <script src="{% static 'js/jquery.datetimepicker.full.min.js' %}"></script>
```
keyvankhademi, line 9: These static files should be in the global static folder. Please put them in static/datetimepicker/[version]/[js|css].
for example static/datetimepicker/2.3.7/js/datetimepicker.js

keyvankhademi (approved):

PR #69, opened 2020-11-03
Feature: submission feedback report

rachellegelden:
## Summary
- added `type` column to Event with options Practice, Assignment, Exam
- reformatted submission details page
- added feedback section that will only be shown for questions that belong to events that are not exams
- added functions for formatting xml results


course/models/models.py
```diff
@@ -378,6 +378,28 @@ def get_decoded_results(self):
...
+
+    def get_failed_test_results(self):
+        all_tests = self.get_decoded_results()
+        return list(filter(get_failed_tests, all_tests))
+
+    def get_num_tests(self):
+        return len(self.get_decoded_results())
+
+
+def get_failed_tests(test):
```
keyvankhademi, line 396: The function should be here in models.py. I would recommend using the lambda function to implicitly define these in the filter.

course/utils/utils.py
```diff
@@ -175,3 +178,18 @@ def create_java_question(pk=None, title=None, text=None, max_submission_allowed=
             event=event,
         )
         question.save()
+
+
+def format_message(message):
```
keyvankhademi, line 183: these function are used locally in junit_xml.py can you please move them there?

keyvankhademi (approved):

PR #81, opened 2020-11-18
Api topics links

Opeyem1a:
- added links as a field in the question-category api


course/migrations/0009_questioncategory_link.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 3.0.7 on 2020-11-18 07:56
```
keyvankhademi, line 1: You can't have more than 1 migration per app per pull request. It happened because you are were testing on your local. You have to delete the migration and make a new one.

Follow these steps to make sure you don't mess up your local database otherwise you have to delete your local database.

1. fake roll back all three migrations (use --fake option)
2. delete the migration files
3. make a new migration (manage.py makemigrations)
4. fake migrate the new migration

api/serializers.py
```diff
@@ -46,9 +46,13 @@ def create(self, validated_data):
 
 
 class QuestionCategorySerializer(serializers.ModelSerializer):
+    linkedTo = serializers.SerializerMethodField('link_to_ids')
     numQues = serializers.SerializerMethodField('count_questions')
```
keyvankhademi, line 50: please rename this to numQuestions.

course/models/models.py
```diff
@@ -24,6 +24,7 @@ class QuestionCategory(models.Model):
     name = models.CharField(max_length=100)
     description = models.TextField()
     parent = models.ForeignKey('self', on_delete=models.SET_NULL, null=True, blank=True)
+    links_to = models.ManyToManyField('self', related_name="linked_by", symmetrical=False, blank=True)
```
keyvankhademi, line 27: the name ('links_to', 'linked_by') gives the impression that it's a foreign key. I recommend using ('links', 'reverse_links) or ('dependents', 'dependees').
Tbh, I don't like either of them. See if you can come up with a better combination.

keyvankhademi (approved):

PR #88, opened 2020-11-25
CG-87: updated text for open event button

rachellegelden:
Instructor View:
2020-11-24 22_35_02-HomePage

Student View:
2020-11-24 22_42_04-HomePage

### DIFF
--- a/canvas/templatetags/canvas.py
+++ b/canvas/templatetags/canvas.py
@@ -35,9 +35,3 @@ def event_button_text(event, user):
         return "Not Available"
     else:
         return "Open"
-
-
-
-
-
-
--- a/canvas/templatetags/canvas.py
+++ b/canvas/templatetags/canvas.py
@@ -15,7 +15,7 @@ def is_allowed_to_open_course(course, user):
 
 @register.simple_tag
 def is_allowed_to_open_event(event, user):
-    return event.has_view_permission(user) or event.status == "Closed"
+    return event.has_view_permission(user) or event.can_view_results(user)
 
 
 @register.simple_tag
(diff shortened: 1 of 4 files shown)


canvas/templatetags/canvas.py
```diff
@@ -15,4 +15,29 @@ def is_allowed_to_open_course(course, user):
 
 @register.simple_tag
 def is_allowed_to_open_event(event, user):
-    return event.has_view_permission(user)
+    return event.has_view_permission(user) or event.status == "Closed"
```
keyvankhademi, line 18: I think we have a is_allowed_to_open function in the event model class. this logic should not be here. You can modify it or create that function if we don't have it and call that function here.

canvas/templatetags/canvas.py
```diff
@@ -15,4 +15,29 @@ def is_allowed_to_open_course(course, user):
...
+    if event.has_edit_permission(user):
+        return "Edit"
+    if event.is_open() and event.course.is_registered(user):
+        if event.type == "EXAM":
+            return "Take Exam"
+        elif event.type == "ASSIGNMENT":
+            return "Complete Assignment"
+        else:
+            return "Start Practice"
+    elif event.status == "Closed" and event.course.is_registered(user):
```
keyvankhademi, line 32: I'm really not a fan of logic in templates. The problem is when you change the behaviour of the event class you modify that class and then somehow you need to remember that you have to modify this file too.

And event.status names can change a lot it's better not to add dependencies on it. I recommend making functions like is_open, is_closed, ... and use those to make a get_button_name function in event class and call that function here.

also when you  implement those functions you can probably modify the status property of even to use those functions.

keyvankhademi (approved):

PR #94, opened 2020-11-28
updated number of attempts for questions

rachellegelden:
Exam questions have 10 attempts
Assignment and practice questions have 100 attempts

### DIFF
--- a/course/models/models.py
+++ b/course/models/models.py
@@ -63,7 +63,7 @@ class Question(PolymorphicModel):
     title = models.CharField(max_length=300, null=True, blank=True)
     text = RichTextField(null=True, blank=True)
     answer = models.TextField(null=True, blank=True)
-    max_submission_allowed = models.IntegerField(default=5, blank=True)
+    max_submission_allowed = models.IntegerField(default=None, blank=True)
     tutorial = RichTextField(null=True, blank=True)
     time_created = models.DateTimeField(auto_now_add=True)
     time_modified = models.DateTimeField(auto_now=True)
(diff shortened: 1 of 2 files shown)


course/models/models.py
```diff
@@ -101,6 +101,7 @@ def success_rate(self):
         return total_solved / total_tried
 
     def save(self, *args, **kwargs):
+        self.max_submission_allowed = 10 if self.event.type == 'EXAM' else 100
```
keyvankhademi, line 104: This is in the wrong place. You have to specify it in the default attribute of the field. You can pass a function to it too (instead of a static value).

rachellegelden: That was my first approach, but I wasn't able to use a field in the model to set the default of another field until I had an instance of the model. See https://stackoverflow.com/questions/41343531/django-models-set-default-relative-to-another-field. Unless you know of another way to do it?

I can definitely write a function to set the number of attempts.

keyvankhademi: Oh, I see. The problem is now if we change the max_submission_allowed for a question it will be overwritten and there will be no way for us to change that. What I recommend to do is to remove the default one we already have and in the save function set the default only if the max_submission_allowed is None.

course/models/models.py
```diff
@@ -114,6 +117,9 @@ def has_view_permission(self, user):
     def has_edit_permission(self, user):
         return user.is_teacher
 
+    def get_max_submissions(self, event):
```
keyvankhademi, line 120: This confuses people because it seems like it's a getter got max_submission_allowed. You can even not make it a function as it's just one line. Another thing event. Type can be None that causes an error for all the questions without an event. use `if event is not None and event.type == "EXAM"`.

keyvankhademi (approved):

PR #96, opened 2020-11-29
Update tokens column based on event

rachellegelden:


canvas/templates/canvas/problem_set_snippet.html
```diff
@@ -1,11 +1,17 @@
+{% load canvas %}
+{% tokens_column_name event as tokens_column_name %}
```
keyvankhademi, line 2: can you please move this close to where you used it? It will be hard to understand what tokens_column_name is when we see this in the future.

canvas/templatetags/canvas.py
```diff
@@ -40,3 +40,11 @@ def event_button_text(event, user):
         return "Not Available"
     else:
         return "Open"
+
+
+@register.simple_tag
+def tokens_column_name(event):
+    if event != '' and event.is_exam_and_open():
```
keyvankhademi, line 47: I don't get the event != '' part. I think you have check if the event is an instance of Event.

course/models/models.py
```diff
@@ -249,7 +249,9 @@ def status(self):
 
     @property
     def formatted_current_tokens_received(self):
-        return str(self.tokens_received) + "/" + str(self.question.token_value)
+        is_exam_open = self.question.event.type == 'EXAM' and self.question.event.is_open()
```
keyvankhademi, line 252: You created a is_exam_and_open method you can use it here.

canvas/templatetags/canvas.py
rachellegelden: Good catch, I'll do that

keyvankhademi (approved):

PR #101, opened 2020-12-02
Update grading for exam questions

rachellegelden:
## Summary
- for exam events, the grade for a question comes from the latest submission of the student
- removed all fields/styling from the events page, the event questions page, the submissions page and the question summary page that would indicate to the student how they did on a question. These changes include:
    * on the questions summary the value for the exam event in the `Grade` column is set to `Not Available Yet` if the exam is still open
    * on the events questions page for exam, all styling regarding the colour of the rows (based on whether the student's answer was correct or wrong) is disabled
    * on the question summary page, the `Status` column from the table at the top of the page was removed.
        * **QUESTION**: should this status column be `Submitted`/`Unsubmitted` similar to the `Status` column on the events question page?
    * the `Grade` and `Status` (`Correct` or `Wrong`) columns were removed from the `My Past Submissions` table for questions in exam events
    * after a student submits a MC question, the banner showing if the answer was right or wrong is removed

### DIFF
--- a/course/models/models.py
+++ b/course/models/models.py
@@ -342,12 +342,12 @@ class Submission(PolymorphicModel):
         if not self.finalized:
             self.calculate_grade(commit=False)
 
-        if not self.in_progress and self.is_correct or self.is_partially_correct:
+        if (not self.in_progress and self.is_correct or self.is_partially_correct) or self.question.event.is_exam():
             user_question_junction = self.uqj
             received_tokens = self.grade * get_token_value(self.question.category, self.question.difficulty)
             token_change = received_tokens - user_question_junction.tokens_received
 
-            if token_change > 0:
+            if self.uqj.question.event.is_exam() or token_change > 0:
                 user_question_junction.tokens_received = received_tokens
                 user_question_junction.save()
 
--- a/course/models/models.py
+++ b/course/models/models.py
@@ -252,7 +252,8 @@ class UserQuestionJunction(models.Model):
 
     @property
     def formatted_current_tokens_received(self):
-        return str(self.question.token_value) if self.question.event.is_exam_and_open() \
+        return str(self.question.token_value) \
+            if hasattr(self.question, 'event') and self.question.event.is_exam_and_open() \
             else str(self.tokens_received) + "/" + str(self.question.token_value)
 
     def save(self, **kwargs):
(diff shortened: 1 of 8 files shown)


course/views/multiple_choice.py
```diff
@@ -65,18 +65,21 @@ def _multiple_choice_question_view(request, question, template_name, key):
...
-                    request, messages.SUCCESS,
-                    'Answer submitted. Your answer was correct. You received {} tokens'.format(
-                        round(received_tokens, 2)),
-                )
-            else:
-                messages.add_message(
-                    request, messages.ERROR,
-                    'Answer submitted. Your answer was wrong',
-                )
+            # TODO: after rebasing .is_exam() needs to become .is_exam
```
keyvankhademi, line 68: You can remove the comment now i guess.

course/views/multiple_choice.py
```diff
@@ -65,18 +65,21 @@ def _multiple_choice_question_view(request, question, template_name, key):
...
+            if not question.event.is_exam:
```
keyvankhademi, line 69: what if it's an exam? You might wanna send a message like "Your submission was received".

keyvankhademi: use messages.INFO

course/models/models.py
```diff
@@ -342,12 +342,12 @@ def save(self, *args, **kwargs):
         if not self.finalized:
             self.calculate_grade(commit=False)
 
-        if not self.in_progress and self.is_correct or self.is_partially_correct:
+        if (not self.in_progress and self.is_correct or self.is_partially_correct) or self.question.event.is_exam:
```
keyvankhademi, line 345: You should be really careful about using question.event it might be None. All the questions without an event will face an error here.
also the brackets are wrong, it should be (not self.in_progress) and (self.is_correct or self.is_partially_correct or self.question.event.is_exam)
You don't want to change the grade if its still in progress.

keyvankhademi, line 345: I'm a little worried that you might have used question.event without checking if it's None in other places. can you please check all the usage of question.event and make sure they are used safely?

rachellegelden: For sure, good catch.

keyvankhademi (approved):

PR #103, opened 2020-12-07
Updated event form styling and added type field

rachellegelden:
image
image

### DIFF
--- a/canvas/templates/canvas/event_create.html
+++ b/canvas/templates/canvas/event_create.html
@@ -13,30 +13,36 @@
 {% endblock %}
 
 {% block header %}
-    Create Event for {{ course.name }}
+    {% if header %}
+        {{ header }} Event for {{ event.course.name }}
+    {% else %}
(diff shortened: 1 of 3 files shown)


canvas/templates/canvas/event_create.html
```diff
@@ -13,30 +13,36 @@
 {% endblock %}
 
 {% block header %}
-    Create Event for {{ course.name }}
+    {% if header %}
+        {{ header }} Event for {{ event.course.name }}
+    {% else %}
+        Create Event for {{ course.name }}
```
keyvankhademi, line 19: shouldn't this be {{event.course.name}} ?

rachellegelden: When the template is used to create an event, the event object isn't passed in from the create_event_view (because the event doesn't exist yet), but it does pass in the course object

keyvankhademi (approved):

PR #120, opened 2021-01-07
Duplicate question feature for admin site

aadiraju:


aadiraju:
Some of the commits don't matter as much because I was mistaken on what I was supposed to do so ignore the first two commits if possible

course/admin.py
```diff
@@ -6,8 +6,32 @@
 from course.models.models import Question, VariableQuestion, MultipleChoiceQuestion, Submission, QuestionCategory, \
     CheckboxQuestion, JavaSubmission, JavaQuestion, TokenValue, MultipleChoiceSubmission, UserQuestionJunction
 from course.models.parsons_question import ParsonsQuestion, ParsonsSubmission
+import copy  # Python native copy functionality
```
keyvankhademi, line 9: The comment is unnecessary.

course/admin.py
```diff
@@ -6,8 +6,32 @@
...
 
 
+def duplicate_question(modeladmin, request, queryset):
```
keyvankhademi, line 12: Having an admin action is nice. But what we really need is to have the ability to duplicate questions in the course.
Our TAs/Instructors do not have access to the admin page. The admin page is restricted to the developers.

course/models/models.py
```diff
@@ -162,6 +162,11 @@ def has_view_permission(self, user):
     def has_edit_permission(self, user):
         return user.is_teacher
 
+    def get_duplicate_question(self):
```
keyvankhademi, line 165: You can rename this function to `clone(self)`.
Then you can use question.clone() to duplicate a question.

course/models/models.py
```diff
@@ -162,6 +162,11 @@ def has_view_permission(self, user):
...
+        dupe_q = self
+        dupe_q.pk = None
+        return dupe_q
+
 
 class VariableQuestion(Question):
```
keyvankhademi, line 171: here you should override the clone method to use the inherited clone method and additionally set the question_ptr_id to None. In this way, each question can have its own clone method that works with its inheritance structure even if it changes in the future.

course/admin.py
```diff
@@ -6,8 +6,32 @@
...
+    '''
+    Django Admin Action that will duplicate selected question instances and set their author to the currently logged-in user
+    '''
+    # qitem refers to the instances of the selected questions
+    for qitem in queryset:
+        dupe_q = copy.deepcopy(qitem)
```
keyvankhademi, line 18: here you can use qitem.clone().
Also qitem and dupe_q are bad names try thinking of a better name using python name conventions.
https://realpython.com/python-pep8/#how-to-choose-names

aadiraju:
I added an extra button on the problem set page for TAs and teachers to clone questions, but can only do so individually. We'd have to implement a checkbox type thing for them to delete or clone multiple questions...

PR #123, opened 2021-01-31
APIs: migrate homepage

Opeyem1a:


api/views.py
```diff
@@ -37,3 +38,25 @@ class QuestionCategoryViewSet(viewsets.ReadOnlyModelViewSet):
 class UserStatsViewSet(viewsets.ReadOnlyModelViewSet):
     queryset = MyUser.objects.all()
     serializer_class = UserStatsSerializer
+
+
+class ActionsViewSet(viewsets.ReadOnlyModelViewSet):
```
keyvankhademi, line 43: you need to set up permissions for this view set. Unauthenticated users should receive 401.

api/views.py
```diff
@@ -37,3 +38,25 @@ class QuestionCategoryViewSet(viewsets.ReadOnlyModelViewSet):
...
+    serializer_class = ActionsSerializer
+
+    def get_queryset(self):
+        user = self.request.user
+        is_recent = self.request.query_params.get('recent', False)
```
keyvankhademi, line 48: this parameter is not defined anywhere? I'm 99% sure that it doesn't show up in the documentation. Make sure you have used the proper format of the parameters here according to the django-restframework documentation.

api/views.py
```diff
@@ -37,3 +38,25 @@ class QuestionCategoryViewSet(viewsets.ReadOnlyModelViewSet):
...
+        if is_recent:
+            return user.actions.all().order_by("-time_modified")
+        return user.actions.all()
+
+
+class UQJViewSet(viewsets.ReadOnlyModelViewSet):
```
keyvankhademi, line 54: The same concepts apply here too.

api/views.py
```diff
@@ -37,3 +39,31 @@ class QuestionCategoryViewSet(viewsets.ReadOnlyModelViewSet):
...
+    ?recent - if true, renders the list in recent first order
+    """
+    serializer_class = ActionsSerializer
+    permission_classes = [IsAuthenticated, ]
+
+    def get_queryset(self):
+        user = self.request.user
+        is_recent = self.request.query_params.get('recent', False)
+        if is_recent:
+            return user.actions.all().order_by("-time_modified")
```
keyvankhademi, line 56: Ok i just realized that you are not limiting the number of useractions returned
if you are just ordering them i recommend using OrderFilter
https://www.django-rest-framework.org/api-guide/filtering/#orderingfilter
you can also add a limit to the number of user actions returned
instead of filtering it in angular it should be filtered in the back-end for the sake of performance

api/views.py
```diff
@@ -37,3 +39,31 @@ class QuestionCategoryViewSet(viewsets.ReadOnlyModelViewSet):
...
+
+class UQJViewSet(viewsets.ReadOnlyModelViewSet):
+    serializer_class = UQJSerializer
+    permission_classes = [IsAuthenticated, ]
+
+    def get_queryset(self):
+        user = self.request.user
+        is_recent = self.request.query_params.get('recent', False)
+        if is_recent:
+            return user.question_junctions.all().order_by('-last_viewed')
```
keyvankhademi, line 68: Same concept applies here.

keyvankhademi (approved):

PR #128, opened 2021-02-04
Tokenvalue migration

aadiraju:


keyvankhademi:
Before this branch can be merged. All the tests should pass and there should be no conflicts.
What you need to do is to pull from master and resolve the conflicts.
You can also local run the command `flake8` to see the errors.

api/views.py
```diff
@@ -34,6 +34,12 @@ class QuestionCategoryViewSet(viewsets.ReadOnlyModelViewSet):
     serializer_class = QuestionCategorySerializer
 
 
+class TokenValueViewSet(viewsets.ModelViewSet, mixins.UpdateModelMixin):
+    queryset = TokenValue.objects.all()
```
keyvankhademi, line 38: Here you need to make sure that all the TokenValue objects exist to prevent getting Null for some category-difficulty pair.

course/admin.py
```diff
@@ -8,6 +8,22 @@
 from course.models.parsons_question import ParsonsQuestion, ParsonsSubmission
 
 
+def clone_question(modeladmin, request, queryset):
```
keyvankhademi, line 11: This is not for this PR. Need to be removed.

course/models/models.py
```diff
@@ -162,6 +162,14 @@ def has_view_permission(self, user):
     def has_edit_permission(self, user):
         return user.is_teacher
 
+    def clone(self):
```
keyvankhademi, line 165: This is not for this PR. Need to be removed.

course/templates/problem_set.html
```diff
@@ -90,6 +90,11 @@
                                 <i class="fa fa-pencil-alt" style="color: black"></i>
                             </a>
                         </div>
+                        <div class="action-icon mx-1" style="background-color: plum">
```
keyvankhademi, line 93: This is not for this PR. Need to be removed.

course/urls.py
```diff
@@ -11,6 +11,7 @@
     path('new-problem/parsons', parsons_question_create_view, name='new_problem_parsons'),
     path('submission/<int:pk>', submission_detail_view, name='submission_detail'),
     path('question/<int:pk>/delete', question_delete_view, name='question_delete'),
+    path('question/<int:pk>/clone', question_clone_view, name='question_clone'),
```
keyvankhademi, line 14: This is not for this PR. Need to be removed.

course/views/views.py
```diff
@@ -110,6 +110,21 @@ def question_delete_view(request, pk):
 
     return HttpResponseRedirect(reverse_lazy('course:problem_set'))
 
+@user_passes_test(teacher_check)
+def question_clone_view(request, pk):
```
keyvankhademi, line 114: This is not for this PR. Need to be removed.

api/views.py
aadiraju: So after looking into the objects, it seems like there aren't any null values in general since when a Tokenvalue object is saved, it is always assigned a default value. There isn't any Null values in the Angular side anymore either. So should I still run a null check and just exclude the ones in the queryset that are null?

keyvankhademi (approved):

PR #141, opened 2021-03-16
Refactor token use api

Opeyem1a:
Converting TokenUse api to use a class-based approach


api/views/token_use.py
```diff
@@ -1,37 +1,25 @@
-import re
-
 from django.shortcuts import get_object_or_404
```
keyvankhademi, line 1: you imported the wrong get_object_or_404, please use the rest_framework version for APIs.

api/serializers/course.py
```diff
@@ -1,7 +1,8 @@
 from rest_framework import serializers
 
 from accounts.models import MyAnonymousUser
-from api.serializers import QuestionSerializer, UQJSerializer
+from api.serializers.question import QuestionSerializer
```
keyvankhademi, line 4: Either fix all of them or leave them as they are. also the same for __init__.py s. I would recommend to fix them all if you want :D

Opeyem1a: fixing as in making it so every import come from the specific file it references?

keyvankhademi: No, its importing all the serializers in __init__.py and having one line of import here.

api/views/action.py
```diff
@@ -2,7 +2,7 @@
 from rest_framework.permissions import IsAuthenticated
 
 from api.pagination import BasePagination
-from api.serializers import ActionsSerializer
+from api.serializers.action import ActionsSerializer
```
keyvankhademi, line 5: I think this should be rolled back to what it was.

keyvankhademi (approved):

Issue #161, opened 2021-04-13
Update user stats API


Opeyem1a:
Improve the way we retrieve user stats

PR #162, opened 2021-04-13
Verify Events + User Stats

### LINKED ISSUE #160: Verify Events
Verify access to events based on user and course

Opeyem1a:
+ Ability to verify if a user can access an event from a specific course (I've noted the changes that need to be made here from our merge conflict talk)
+ Updated the way we retrieve user success rate
Closes #160, #161


api/views/course.py
```diff
@@ -163,3 +163,49 @@ def verify(self, request, pk=None):
...
             "attempts_remaining": course_reg.verification_attempts,
         })
+
+    @action(detail=True, methods=['get'], url_path='validate-event/(?P<event_pk>[^/.]+)')
+    def validate_event(self, request, pk=None, event_pk=None):
+        """
+        Validates that an event belongs to a particular course.
+        """
+        course = get_object_or_404(CanvasCourse, pk=pk)
+        registered = self.request.query_params.get('registered', None)
```
keyvankhademi, line 173: I don't understand why this "registered" is necessary. I think we can remove it completely and nothing bad happens.

Opeyem1a: the goal is that if the user manually types in "/course/1/event/2" then we check if event 2 is in course 1 and also check if they're even registered in course 1, to begin with. The reason I have 'registered' as a boolean, is because I presumed there might be a need for admin users to access events within courses that they aren't registered in.

Although based on our other discussions, I realize this is a pretty sloppy way to handle this - I need to refactor every place that uses this logic to simply check if request.user is an admin or not

keyvankhademi (approved):

PR #176, opened 2021-05-20
Only Author Can Delete A Question

### LINKED ISSUE #175: Only Author can Delete Questions

carson-ricca:
- Tested functionality and it works. Only the author of the question is allowed to delete it.
- Closes #175


api/permissions.py
```diff
@@ -17,3 +17,10 @@ def has_permission(self, request, view):
 
     def has_object_permission(self, request, view, obj):
         return request.user == obj.user
+
+
+class IsOwnerOrReadOnly(permissions.IsAuthenticated):
```
keyvankhademi, line 22: We want them to be able to edit it but not delete it.

keyvankhademi:
This also has a UI task to disable the delete button if they are not the author.

api/permissions.py
carson-ricca: Oh whoops I forgot this disabled that.

carson-ricca:
> This also has a UI task to disable the delete button if they are not the author.

Will fix this tomorrow.

keyvankhademi (approved):

PR #185, opened 2021-05-31
Add is_author field to Questions

### LINKED ISSUE #184: Add if User is Author to Question

carson-ricca:
Closes #184

### DIFF
--- a/api/serializers/question.py
+++ b/api/serializers/question.py
@@ -23,8 +23,7 @@ class QuestionSerializer(serializers.ModelSerializer):
         return request.user.question_junctions.get(question__pk=obj.id).status
 
     def get_is_author(self, obj):
-        print(obj.author)
         request = self.context.get('request', None)
         if request is None:
-            return ''
+            return False
         return request.user == obj.author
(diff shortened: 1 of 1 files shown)


api/serializers/question.py
```diff
@@ -7,16 +7,24 @@
...
+                  'is_open', 'is_exam', 'is_exam_and_open', 'status', 'full_category_name', 'is_author']
 
     def get_uqj_status(self, obj):
         request = self.context.get('request', None)
         if request is None:
             return ''
         return request.user.question_junctions.get(question__pk=obj.id).status
+
+    def get_is_author(self, obj):
+        print(obj.author)
```
keyvankhademi, line 26: remove the log, please.

api/serializers/question.py
```diff
@@ -7,16 +7,24 @@
...
+        request = self.context.get('request', None)
+        if request is None:
+            return ''
```
keyvankhademi, line 29: return type should be consistent. 
```suggestion
            return False
```

api/serializers/question.py
keyvankhademi:
```suggestion
```

PR #190, opened 2021-06-04
Add Support for Checkbox Questions

### LINKED ISSUE #186: Option for MCQ to be a Checkbox Question

carson-ricca:
Closes #186


course/models/models.py
```diff
@@ -456,6 +459,13 @@ def submit(self):
...
     @property
     def answer_display(self):
+        if self.uqj.is_checkbox():
+            values = []
+            for key, value in self.uqj.get_rendered_choices().items():
+                for answer in self.answer.split(','):
+                    if key == answer:
+                        values.append(value)
+            return values
         return self.uqj.get_rendered_choices().get(self.answer, 'Unknown')
```
keyvankhademi, line 469:
```suggestion
    @property
    def answer_display(self):
        values = []
        rendered_choices = self.uqj.get_rendered_choices()
        for answer in self.answer.split(','):
            values.append(rendered_choices.get(answer, 'Unknown'))
        return values  
```
We can't have a function that returns 2 different types of values.
Also, I'm not a fan of nested for loops in this case.
It also ignores the answers that are not in the choices. We should somehow specify something weird has happened.

course/views/views.py
```diff
@@ -69,9 +58,6 @@ def question_view(request, pk, key=None):
     if isinstance(question, JavaQuestion):
         return _java_question_view(request, question, key)
 
-    if isinstance(question, CheckboxQuestion):
-        return _multiple_choice_question_view(request, question, 'checkbox_question.html', key)
```
keyvankhademi, line 73: The html files that are not being used anymore should also be deleted.

keyvankhademi (changes requested):

course/views/views.py
carson-ricca: Do I delete just the html files or do I delete all the rendering related to problems?

course/grader/grader.py
```diff
@@ -17,14 +17,19 @@ def grade(self, submission):
 class MultipleChoiceGrader(Grader):
 
     def grade(self, submission):
-        if submission.answer != submission.question.answer:
-            return False, 0
+        correct_count = 0
```
carson-ricca, line 20: A basic implementation that supports the grading of checkbox questions. Needs requirement clarifications then improvements can be made.

keyvankhademi (approved):

PR #209, opened 2021-07-02
Remove Practice Events

### LINKED ISSUE #208: Remove Practice Event Types

carson-ricca:
Closes #208


canvas/migrations/0009_auto_20210702_1305.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('canvas', '0008_event_type'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='event',
+            name='type',
+            field=models.CharField(choices=[('ASSIGNMENT', 'ASSIGNMENT'), ('EXAM', 'EXAM')], max_length=500),
```
keyvankhademi, line 16: This will give us an error if there is already a practice event in the database. An idea is to change them to assignments before this database change.
Make sure to test this properly by adding a practise event and running the migration

keyvankhademi (changes requested):

canvas/migrations/0009_auto_20210702_1305.py
carson-ricca: Tested with a practice event locally and it was changed to assignment on the migration.

keyvankhademi (approved):

PR #213, opened 2021-07-09
Deleted Question gets Flagged

### LINKED ISSUE #216: Deleted Questions get Flagged

carson-ricca:
Closes #216


course/models/models.py
```diff
@@ -111,6 +111,15 @@ class Question(PolymorphicModel):
 
     is_verified = models.BooleanField(default=False)
 
+    CREATED = 'CRE'
```
keyvankhademi, line 114: Can you please rename this to complete CREATED and DELETED?

course/models/models.py
```diff
@@ -175,6 +184,10 @@ def save(self, *args, **kwargs):
         super().save(*args, **kwargs)
         ensure_uqj(None, self)
 
+    def delete(self):
```
keyvankhademi, line 187: This is bad! We won't be able to delete questions from admin either!
This logic should be in the API only.

keyvankhademi (changes requested):

course/models/models.py
carson-ricca: Whoops right!

course/migrations/0020_merge_20210721_1308.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 3.0.14 on 2021-07-21 20:08
```
carson-ricca, line 1: Merged and update model migration. Didn't want to delete and attempt to recreate as it might have conflicted with all the migrations from master.

keyvankhademi (approved):

PR #215, opened 2021-07-12
Integration

JarodAlgra:


api/views/team.py
```diff
@@ -0,0 +1,128 @@
...
+        # get the team name from the request
+        team_name = request.data.get("name", None)
+
+        course = get_object_or_404(CanvasCourse, pk=pk)
+
+        # all teams registrations for the user in the requested course
+        all_team_reg = TeamRegistration.objects.filter(user=request.user, team__course=course)
+        
+        # delete other registrations if any exist
+        if(all_team_reg is not None):
```
carson-ricca, line 64: Shouldn't there only ever be at most a single `team_reg` for each user? If this is the case instead of adding and deleting `TeamRegistration` objects could we update the `Team` value in the User's `TeamRegistration` object after creating the new team? This way we aren't making a ton of delete and creates to the `TeamRegistration` model. Only if they don't have an existing `team_reg` then we can create a new one.

api/views/team.py
```diff
@@ -0,0 +1,128 @@
...
+        # create a new team with the specified name
+        new_team = Team(name=team_name, course=course)
+        new_team.save()
+
+        # register the creator of the team to the team
+        new_team_reg = TeamRegistration(user=request.user, team=new_team)
+        new_team_reg.save()
+        
+
+        return Response({
```
carson-ricca, line 78: Trying to remove the usage of the status response in our APIs. Instead, you can return the created object, or nothing and just a response code, then in the front-end do what is required based on the response instead of the message in `status`.

api/views/team.py
```diff
@@ -0,0 +1,128 @@
...
+        course = team.course
+        
+        # all teams registrations for the user in the requested course
+        all_team_reg = TeamRegistration.objects.filter(user=request.user, team__course=course)
+        
+
+        # if the user is not registered for the selected team
+        if team_reg is None:
+            # delete other registrations if any exist already
+            if(all_team_reg is not None):
```
carson-ricca, line 99: Similar to above, can update their registration instead of deleting and re-creating. If this will result in just updating the user's team-registration than this can probably be an `update` (`PUT`)

api/views/team.py
```diff
@@ -0,0 +1,128 @@
...
+                for reg in all_team_reg:
+                    reg.delete()
+
+            # create a TeamRegistration for this (user, team) pair if one does not already exist
+            team_reg = TeamRegistration(user=request.user, team=team)
+            team_reg.save()
+
+        return Response({
+                "status": "Registered",
```
carson-ricca, line 108: Similar to the response comment above.

api/views/team.py
```diff
@@ -0,0 +1,128 @@
...
+            })
+        
+
+    @action(detail=True, methods=['post'])
+    def leave_team(self, request, pk=None):
```
carson-ricca, line 113: This could probably be a `delete` method. Then you can just override the `destroy` method and set the user's `team_reg` to `None` instead of completely deleting the object. Look at the `destroy` function in the question viewset. Changing this to override the `destroy` method means you could then make a `delete` request from the front-end.

api/views/team.py
```diff
@@ -0,0 +1,128 @@
...
+        team_reg = get_team_registration(user=request.user, team=team)
+
+        if team_reg is None:
+            pass
+        
+        else:
+            team_reg.delete()
+
+        return Response({
+                "status": "De-registered",
```
carson-ricca, line 127: Same as above response comments.

canvas/admin.py
```diff
@@ -1,10 +1,13 @@
+
```
carson-ricca, line 1: Extra line.

canvas/migrations/0010_team_teamregistration.py
```diff
@@ -0,0 +1,38 @@
+# Generated by Django 3.0.14 on 2021-07-12 18:06
```
carson-ricca, line 1: This and the above migration should be combined into a single one. When creating new features want just a single migration, that includes all the changes to the models.

canvas/models.py
```diff
@@ -11,7 +11,7 @@
 
 class CanvasCourse(models.Model):
     mock = models.BooleanField(default=False)
-    name = models.CharField(max_length=500)
+    name = models.CharField(max_length=500, unique=True)
```
carson-ricca, line 14: The reason behind changing this?

canvas/templates/canvas/course_list.html
```diff
@@ -33,7 +33,6 @@
                 </td>
                 <td>
                     {% if allowed_to_open %}
-                        <a href="{% url 'canvas:course' course.pk %}" class="btn btn-primary">Open</a>
```
carson-ricca, line 36: Can be reverted.

canvas/utils/utils.py
```diff
@@ -25,3 +25,9 @@ def get_course_registration(user, course):
 
     qs = CanvasCourseRegistration.objects.filter(user=user, course=course)
     return qs.get() if qs.exists() else None
+
+def get_team_registration(user, team):
+    from canvas.models import TeamRegistration
+
+    qs = TeamRegistration.objects.filter(user=user, team=team)
```
carson-ricca, line 32: I believe you can just return the results from the filter. It should return `None` if no objects exist that match the parameters.

canvas_gamification/urls.py
```diff
@@ -39,6 +39,9 @@
         path('api-auth/', include('rest_framework.urls', namespace='rest_framework')),
         path('api/', include('api.urls', namespace='api')),
         path('canvas/', include('canvas.urls', namespace='canvas')),
+
+        # path('leaderboard/', include(('leaderboard.urls', 'leaderboard'))),
```
carson-ricca, line 43: Remove comment.

course/models/models.py
```diff
@@ -23,7 +23,6 @@
     ("HARD", "HARD"),
 ]
 
-
```
carson-ricca, line 26: Revert.

course/migrations/0025_auto_20210719_2232.py
```diff
@@ -0,0 +1,38 @@
+# Generated by Django 3.0.14 on 2021-07-20 05:32
```
carson-ricca, line 1: This and all the above migrations that were added should be combined into a single migration. Also, note you will want to be up-to-date with master as more migrations have happened since then. Will help ensure the version numbering of the migrations is correct. Same as above we want a new feature implementation to only have a single migration.

carson-ricca (changes requested):

api/views/team.py
JarodAlgra: We were thinking there would be one team_reg per user per course
there should only be one per user per course though, 
so we should be able to just update the specific registration

.gitignore
```diff
@@ -6,7 +6,6 @@ debug.log
 .idea/
 env/gamification.env
 
-
```
carson-ricca, line 9: Change can be ignored.

api/views/team.py
```diff
@@ -0,0 +1,122 @@
...
+        try:
+            team_reg = TeamRegistration.objects.get(user=user, team__course=course)
+        except ObjectDoesNotExist:
+            return Response({
+                'user_id': user.id,
+                'team_id': -1
+            })
+
+        return Response({
+            'user_id': team_reg.user.id,
```
carson-ricca, line 37: Either return the object itself or return an HTTP response code.

api/views/team.py
```diff
@@ -0,0 +1,122 @@
...
+        # if the user does not have a team registration
+        # create one
+        else:
+            # register the creator of the team to the team
+            new_team_reg = TeamRegistration(user=request.user, team=new_team)
+            new_team_reg.save()
+
+        return Response({
+            200
+        })
```
carson-ricca, line 73: Update this to use the response HTTP. Something like 
```suggestion
        return Response(status=status.HTTP_201_CREATED)
```

api/views/team.py
```diff
@@ -0,0 +1,122 @@
...
+        # if the user is not registered in any team in this course
+        # create a team_registration
+        if not team_reg.exists():
+            # create a TeamRegistration for this (user, team) pair
+            team_reg = TeamRegistration(user=request.user, team=team)
+            team_reg.save()
+
+            return Response({
+                200
+            })
```
carson-ricca, line 97: Note the comment about responses above.

api/views/team.py
```diff
@@ -0,0 +1,122 @@
...
+
+        # if the user is/was registered in a team 
+        # update their team_registration
+        else:
+            team_reg.update(team=team)
+
+            return Response({
+                'status': 200
+            })
```
carson-ricca, line 106: Same comment as above ^

api/views/team.py
```diff
@@ -0,0 +1,122 @@
...
+        """
+        team = get_object_or_404(Team, pk=kwargs['pk'])
+        team_reg = get_team_registration(user=request.user, team=team)
+
+        if team_reg is None:
+            pass
+
+        else:
+            team_reg.team = None
+            team_reg.save()
```
carson-ricca, line 120: This can be restructured to be simpler.

api/views/team.py
```diff
@@ -0,0 +1,122 @@
...
+
+        return Response({200})
```
carson-ricca, line 122: Response as stated above.

canvas/migrations/0010_merge_20210730_1543.py
```diff
@@ -0,0 +1,14 @@
+# Generated by Django 3.0.14 on 2021-07-30 22:43
```
carson-ricca, line 1: Can this and the above migration be combined after merging with the latest version on master.

canvas/models.py
```diff
@@ -315,3 +315,38 @@ def revert(self):
...
+        if user.is_anonymous:
+            return False
+        return self.teamregistration_set.filter(user=user).exists()
+
+    @property
+    def tokens(self):
+        token_count = 0
+        for registration in self.teamregistration_set.all():
+            if registration.total_tokens_received:
+                token_count += int(registration.total_tokens_received)
```
carson-ricca, line 337: Are you sure we wanted to cast this value to an int?

canvas/templates/canvas/course_list.html
carson-ricca, line 36: Reason for this removal?

static/angular/9-es2015.28d9eb8bf586c5747678.js
```diff
@@ -0,0 +1 @@
...rtSegment(r),t},removeSegment:function(t){var e=this.segments,n=e.length;if(0===n)throw new Error("Path has no segments.");if(t<0&&(t=n+t),t>=n||t<0)throw new Error("Index out of range.");var r=e.splice(t,1)[0],i=r.previousSegment,o=r.nextSegment;i&&(i.nextSegment=o),o&&(o.previousSegment=i),r.isSubpathStart&&o&&this.updateSubpathStartSegment(o)},replaceSegment:function(t,e){var n,r=this.segments,i=r.length;if(0===i)throw new Error("Path has no segments.");if(t<0&&(t=i+t),t>=i||t<0)throw new Error("Index out of range.");var o=r[t],s=o.previousSegment,a=o.nextSegment,c=o.isSubpathStart;if(Array.isArray(e)){if(!(e=e.reduce(function(t,e){return t.concat(e)},[]))[0].isSegment)throw new Error("Segments required.");r.splice(t,1);for(var u=e.length,l=0;l<u;l++)n=this.prepareSegment(e[l],s,a),r.splice(t+l,0,n),s=n,c&&n.isSubpathStart&&(c=!1)}else{if(!e||!e.isSegment)throw new Error("Segment required.");n=this.prepareSegment(e,s,a),r.splice(t,1,n),c&&n.isSubpathStart&&(c=!1)}c&&a&&this.updateSubpathStartSegment(a)},round:function(t){for(var e=this.segments,n=e.length,r=0;r<n;r++)e[r].round(t);return this},scale:function(t,e,n){for(var r=this.segments,i=r.length,o=0;o<i;o++)r[o].scale(t,e,n)
```
carson-ricca, line 1: Not sure why this is added?

static/angular/9-es5.28d9eb8bf586c5747678.js
```diff
@@ -0,0 +1 @@
...rentiable())return!0;return!1},isValid:function(){var t=this.segments;return 0===t.length||"M"===t[0].type},length:function(t){var e=this.segments,n=e.length;if(0===n)return 0;for(var r=void 0===(t=t||{}).segmentSubdivisions?this.getSegmentSubdivisions({precision:void 0===t.precision?this.PRECISION:t.precision}):t.segmentSubdivisions,i=0,o=0;o<n;o++)i+=e[o].length({subdivisions:r[o]});return i},lengthAtT:function(t,e){var n=this.segments,r=n.length;if(0===r)return 0;var i=t.segmentIndex;if(i<0)return 0;var o=t.value;i>=r?(i=r-1,o=1):o<0?o=0:o>1&&(o=1);for(var a=void 0===(e=e||{}).precision?this.PRECISION:e.precision,s=void 0===e.segmentSubdivisions?this.getSegmentSubdivisions({precision:a}):e.segmentSubdivisions,u=0,c=0;c<i;c++){var l=n[c];u+=l.length({precisison:a,subdivisions:s[c]})}return u+(l=n[i]).lengthAtT(o,{precisison:a,subdivisions:s[i]})},pointAt:function(t,e){if(0===this.segments.length)return null;if(t<=0)return this.start.clone();if(t>=1)return this.end.clone();var n=void 0===(e=e||{}).precision?this.PRECISION:e.precision,r={precision:n,segmentSubdivisions:void 0===e.segmentSubdivisions?this.getSegmentSubdivisions({precision:n}):e.segmentSubdivisions},i=this.length(r);
```
carson-ricca: Also not sure why this is added.

carson-ricca (changes requested):
Remember to combine all the migrations into a singular migration that is up to date with the master branch.

PR #219, opened 2021-07-16
Import Events

### LINKED ISSUE #214: Course Events Import Event

carson-ricca:
Closes #214


api/views/event.py
```diff
@@ -28,3 +29,27 @@ def get_event_types(self, request, pk=None):
         Returns a dictionary of the defined event types
         """
         return Response(EVENT_TYPE_CHOICES)
+
+    @action(detail=False, methods=['post'], url_path="duplicate-event")
+    def duplicate_event(self, request):
+        """
+        Duplicates an event as well as the questions within the event.
+        """
+        event = Event.objects.filter(id=request.data.get("event", None)).first()
```
keyvankhademi, line 38: We have to check if they exist first and respond accordingly with 404 and the correct message if they don't.

api/views/event.py
```diff
@@ -28,3 +29,27 @@ def get_event_types(self, request, pk=None):
...
+        course = CanvasCourse.objects.filter(id=request.data.get("course", None)).first()
+        old_event_id = event.id
+        event.id = None
```
keyvankhademi, line 41: Let's add a function to the event called `clone` and here you just call `event.clone()`.
Similarly you can have `question.clone()` and use it in the event clone function.

api/views/event.py
```diff
@@ -28,3 +29,27 @@ def get_event_types(self, request, pk=None):
...
+        event.course = course
+        event.name += ' (Copy)'
+        event.save()
+        for question in Question.objects.all().filter(event=old_event_id):
+            question.id = None
+            question.pk = None
+            question.question_ptr_id = None
```
keyvankhademi, line 48: I'm like 99.9% sure that we need to set the `question.variablequestion_ptr_id = None` too. But maybe because `VariableQuestion` is a child of `Question` this could work. But let's be safe and do it properly.

api/views/event.py
```diff
@@ -28,3 +29,27 @@ def get_event_types(self, request, pk=None):
...
+            question.course = course
+            question.event = event
+            question.title += ' (Copy)'
+            question.save()
+        return Response({
+            "success": True,
```
keyvankhademi, line 54: We are deprecating the `success` we should check the response status instead. Here i think you can either return the new event or nothing.

keyvankhademi (changes requested):

api/views/event.py
carson-ricca: Okay I will set this as well.

course/models/models.py
```diff
@@ -185,6 +185,19 @@ def has_view_permission(self, user):
     def has_edit_permission(self, user):
         return user.is_teacher
 
+    def clone(self, course, event, author):
+        question_clone = self
```
keyvankhademi, line 189: This will modify the current instance I think!
You can confirm it by getting a question and then clone it and check if the original one is changed or not.
Please check if it modifies the original one you have to change this to get a copy of `self`, not the `self` itself.

canvas/models.py
```diff
@@ -269,6 +269,14 @@ def cannot_access_event_yet(self, user):
     def is_exam_and_open(self):
         return self.is_exam and self.is_open
 
+    def clone(self, course):
+        course_clone = self
```
keyvankhademi, line 273: Here you should iterate over the questions and clone them. The clone method should work independently.

api/views/event.py
```diff
@@ -28,3 +29,22 @@ def get_event_types(self, request, pk=None):
...
+        event = Event.objects.filter(id=request.data.get("event", None)).first()
+        course = CanvasCourse.objects.filter(id=request.data.get("course", None)).first()
+        current_author = None
+        if request.user.is_authenticated:
+            current_author = request.user
+        if event and course:
+            old_event_id = event.id
+            event.clone(course)
+            for question in Question.objects.all().filter(event=old_event_id):
+                question.clone(course, event, current_author)
```
keyvankhademi, line 47: This logic moves to `event.clone()`. (See the other comment)

api/views/event.py
keyvankhademi: I think this is not resolved.

keyvankhademi (changes requested):

course/models/models.py
carson-ricca: I have tested and this does not modify the current instance. Once the `id` and respective fields are set to `None`, a new instance is created and the new instance is the one being modified.

api/views/event.py
carson-ricca: Kinda confused why. If I make a request with an invalid course or event through postman, I get a 404 error with the defined error message.
Screen Shot 2021-07-22 at 3 30 12 PM

course/models/models.py
keyvankhademi: question clone

Cloning from q1 modifies the q1 but as long as it's not saved the changes won't appear in the database.

api/views/event.py
keyvankhademi: Ok yes, I see you did the validation at the end of the function. I think it should be at the beginning of the function. You can use `get_object_or_404` shortcut.

carson-ricca: Okay yeah that makes sense!

course/models/models.py
carson-ricca: I've attached a video. As you can see in the demo the original questions and events are not modified with the current code when importing an event.
https://user-images.githubusercontent.com/59884796/127220138-2b0ea408-e81b-4de0-beac-eca4688719a8.mp4

keyvankhademi (approved):

PR #234, opened 2021-09-04
Support Variables in Input File Names

### LINKED ISSUE #233: Add Variables to File Names (Java Question)

carson-ricca:
- Closes #233


course/models/models.py
```diff
@@ -300,13 +300,16 @@ def get_lines(self):
             ]
             random.shuffle(lines)
             rendered_lines.append({
-                'name': input_files['name'],
+                'name': render_text(input_files['name'], self.get_variables()),
```
keyvankhademi, line 303: Have tried submitting a code solution with this?
I think we should do a little more than this. Probably we have to change some stuff if the graders.

carson-ricca: I have to double-check... I think I tested and it worked but I'll make sure.

PR #245, opened 2021-10-06
Added Course Admin API to support course dashboard in the frontend.

shiro102:


api/views/course_admin.py
```diff
@@ -40,6 +43,28 @@ def register_user(self, request, pk=None):
...
+        status_type = request.data.get("type")
+        course_registration = get_object_or_404(CanvasCourseRegistration, id=registration_id)
+
+        if status_type == 1:
+            course_registration.is_blocked = status
+            course_registration.save()
+
+        if status_type == 2:
+            course_registration.is_verified = status
+            course_registration.save()
```
aadiraju, line 66: What exactly is `status_type` here? If it is arbitrary, then maybe make it descriptive, instead of just numbers which aren't readable

shiro102: I changed that. Is it appropriate now?

api/views/course_admin.py
```diff
@@ -16,11 +17,13 @@ class CourseAdminViewSet(viewsets.GenericViewSet):
     permission_classes = [TeacherAccessPermission, ]
     queryset = CanvasCourse.objects.all()
     serializer_class = CourseSerializer
+    search_fields = ['user__first_name', 'user__last_name', 'user__email']
```
keyvankhademi, line 20: There is a logical problem here. Notice that the view set is for the model `CanvasCourse` and has a `CourseSerializer`. Now you added search fields for `user__first_name` and ... which are properties of `CanvasCourseRegistration` and not `CanvasCourse`. Then you are manually calling the filter for a `CanvasCourseRegistration` queryset. That is why it works for that specific action. But it will break the general functionality of the view set.

api/views/course_admin.py
```diff
@@ -40,6 +43,28 @@ def register_user(self, request, pk=None):
         if course.is_registered(user):
             raise ValidationError(ERROR_MESSAGES.COURSE_REGISTRATION.ALREADY_REGISTERED)
 
-        # TODO: Implement the registration
+        # TODO: Implement the registration that uses of canvas_user_id which can be done from canvas api
 
         return Response(request.data)
+
+    @action(detail=True, methods=['post'], url_path='change-status')
+    def change_status(self, request, pk=None):
```
keyvankhademi, line 51: This is not the way we want to implement it. It just a matter of project project convention there is nothing wrong with what you did. Please rename this to `update_registration` which gets 2 fields `is_verified` and `is_blocked` then it will update them. Then is the front-end you'll have only 1 function in the service for it and both button will you that service.

api/serializers/canvas_course_registration.py
```diff
@@ -29,4 +29,4 @@ def get_is_registered(self, course):
     class Meta:
         model = CanvasCourseRegistration
         fields = ['id', 'canvas_user_id', 'is_verified', 'is_blocked', 'token_uses', 'total_tokens_received',
-                  'available_tokens', 'username', 'name']
```
keyvankhademi, line 32: I believe these changes are already in master. Please pull from master and avoid duplicated code.

canvas/models.py
```diff
@@ -214,8 +214,16 @@ def username(self):
         return self.user.username
 
     @property
-    def name(self):
-        return self.user.get_full_name()
+    def first_name(self):
```
keyvankhademi, line 217: We don't need to expose first_name and last_name separately. And as I mentioned before email is an internal property we front-end only works with the username. Please remove the email.

api/views/course_admin.py
```diff
@@ -16,11 +17,13 @@ class CourseAdminViewSet(viewsets.GenericViewSet):
...
     queryset = CanvasCourse.objects.all()
     serializer_class = CourseSerializer
+    search_fields = ['user__first_name', 'user__last_name', 'user__email']
+    filter_backends = [filters.SearchFilter, DjangoFilterBackend]
 
     @action(detail=True, methods=['get'], url_path='registered-users')
     def registered_users(self, request, pk=None):
         course = get_object_or_404(self.get_queryset(), pk=pk)
-        course_registration_list = course.canvascourseregistration_set.all()
+        course_registration_list = self.filter_queryset(course.canvascourseregistration_set.all())
```
keyvankhademi, line 26: Here you need to manually filter the course_registration set. If you don't know how to do it ask Abi to help you.

canvas/models.py
shiro102: Fixed.

api/views/course_admin.py
```diff
@@ -40,6 +49,30 @@ def register_user(self, request, pk=None):
...
+        if status == 'VERIFIED':
+            course_registration.verify()
+        if status == 'BLOCKED':
+            course_registration.block()
+        if status == 'PENDING_VERIFICATION':
+            course_registration.unverify()
+        if status == 'UNREGISTERED':
+            course_registration.unregister()
+
+        course_registration.save()
```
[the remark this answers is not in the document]
shiro102: Done

api/views/course_admin.py
```diff
@@ -16,11 +18,19 @@ class CourseAdminViewSet(viewsets.GenericViewSet):
     permission_classes = [TeacherAccessPermission, ]
     queryset = CanvasCourse.objects.all()
     serializer_class = CourseSerializer
+    ordering_fields = []
+    filter_backends = [filters.OrderingFilter, DjangoFilterBackend]
 
     @action(detail=True, methods=['get'], url_path='registered-users')
     def registered_users(self, request, pk=None):
+        self.ordering_fields = ['id', 'user__username', 'user__first_name', 'status']
```
aadiraju, line 26: This is redundant here, move this up to the class variable declared above

canvas/models.py
```diff
@@ -159,6 +164,30 @@ class Meta:
...
+        self.status = 'BLOCKED'
+        self.save()
+
+    def unregister(self):
+        self.status = 'UNREGISTERED'
+        self.save()
+
+    def unverify(self):
+        self.status = 'PENDING_VERIFICATION'
+        self.save()
```
aadiraju, line 189: Maybe combine into one method, that takes in an input on what to save as. so something like `change_status(self, status)`. Sample call would be like `change_status(STATUS["UNREGISTERED"])`

general/services/action.py
```diff
@@ -254,3 +254,29 @@ def create_submission_evaluation_action(submission):
             'status': submission.status,
         }
     )
+
+
+def course_registration_update_action(registration, user, data):
+    Action.create_action(
+        actor=user,
+        description='User update a course registration status.',
```
aadiraju, line 262: Maybe change it to "User updated course registration status"

aadiraju (commented):
I've added some changes, please address those. Otherwise it looks good!

canvas/models.py
shiro102: I dont think we should do it as I just notice that I have to make STATUS array exported to use it in course_admin or other views.

api/views/course_admin.py
shiro102: Done.

general/services/action.py
shiro102: Done.

api/views/course_admin.py
```diff
@@ -40,6 +49,37 @@ def register_user(self, request, pk=None):
         if course.is_registered(user):
             raise ValidationError(ERROR_MESSAGES.COURSE_REGISTRATION.ALREADY_REGISTERED)
 
-        # TODO: Implement the registration
+        # TODO: Implement the registration that uses of canvas_user_id which can be done from canvas api
+        canvascourseregistration = CanvasCourseRegistration(course=course, user=user)
```
keyvankhademi, line 53: Registering a user without setting the canvas_user_id will break the functionality. Please talk to bowen and figure out a way to do it. My idea is to pass the student id for this and find the canvas user based on the student id.

canvas/models.py
```diff
@@ -182,12 +211,12 @@ def check_verification_code(self, code):
         if self.is_blocked:
             return False
         if str(self.verification_code) == str(code):
-            self.is_verified = True
+            self.verify()
             self.save()
```
keyvankhademi, line 215: we don't need the self.save() here.

canvas/models.py
```diff
@@ -182,12 +211,12 @@ def check_verification_code(self, code):
...
             return True
         self.verification_attempts -= 1
         if self.verification_attempts <= 0:
-            self.is_blocked = True
+            self.block()
         self.save()
```
keyvankhademi, line 220: we also don't need the self.save() here.

keyvankhademi (changes requested):

api/views/course_admin.py
```diff
@@ -16,11 +18,18 @@ class CourseAdminViewSet(viewsets.GenericViewSet):
...
     serializer_class = CourseSerializer
+    ordering_fields = ['id', 'user__username', 'user__first_name', 'status']
+    filter_backends = [filters.OrderingFilter, DjangoFilterBackend]
 
     @action(detail=True, methods=['get'], url_path='registered-users')
     def registered_users(self, request, pk=None):
         course = get_object_or_404(self.get_queryset(), pk=pk)
-        course_registration_list = course.canvascourseregistration_set.all()
+        name = request.query_params.get('search', '')
+        course_registration_list = self.filter_queryset(course.canvascourseregistration_set.all())
```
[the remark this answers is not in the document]
shiro102: I have remove the ordering here and will move it to the front-end.

api/views/course_admin.py
shiro102: Done. Remove the register_user function according to the discussion with Bowen.

keyvankhademi (approved):

PR #248, opened 2021-10-09
Added Switch favorite to help toggle favorite in frontend

BeefyAdmiral:
Added switch favorite feature which is used to toggle "favorite" in CourseDashboard Question Snippet


api/views/uqj.py
```diff
@@ -37,3 +41,15 @@ def get_queryset(self):
                 accessible_uqj.append(temp.question)
         uqj = user.question_junctions.filter(question__in=accessible_uqj)
         return uqj
+
+    @action(detail=True, methods=['post'], url_path='switch-favorite')
+    def switch_favorite(self, request, pk=None):
```
keyvankhademi, line 46: For the convention of the project, we don't have toggle APIs. We have create/update APIs. And also this is a read-only api, you should not update anything here.
I think we don't have a viewset to update UQJs. You should create a viewset to update UQJ, which only allows a user to update is_favorite ONLY for the their UQJs.

canvas/admin.py
```diff
@@ -7,4 +7,4 @@
 admin.site.register(CanvasCourseRegistration)
 admin.site.register(Event)
 admin.site.register(TokenUseOption)
-admin.site.register(TokenUse)
+admin.site.register(TokenUse)
```
keyvankhademi, line 10: Don't randomly change files. Please take a look at your changes every time you make a PR.

canvas/models.py
```diff
@@ -336,4 +336,4 @@ def revert(self):
             course_reg.canvas_user_id: {
                 'posted_grade': 0,
             }
-        })
+        })
```
keyvankhademi, line 339: roll back

course/models/models.py
```diff
@@ -235,6 +235,7 @@ class UserQuestionJunction(models.Model):
 
     is_solved = models.BooleanField(default=False, db_index=True)
     is_partially_solved = models.BooleanField(default=False, db_index=True)
+    is_favourite = models.BooleanField(default=False, db_index=True)
```
keyvankhademi, line 238: Are we filtering the favorite question? If not we don't need an index for it.

keyvankhademi (changes requested):

course/models/models.py
keyvankhademi: This change requires a migration. Please have the migration in the PR.

keyvankhademi (changes requested):

canvas/models.py
BeefyAdmiral: I am not able to understand what is the reason for the rollback here? Do you want me to add an additional line would that do it?

PR #265, opened 2021-11-13
Registration Type

### LINKED ISSUE #266: Add Course Registration Type

carson-ricca:
- Closes #266


carson-ricca:
- One odd thing I'm having trouble figuring out is if I change the `registration_type` in the admin portal, `is_verified` gets set to false. Not sure how this is happening.
- Other questions do we want to set the registration type automatically? It is currently set to STUDENT by default but there may be a way to get the correct type automatically everytime?

canvas/models.py
```diff
@@ -128,7 +130,12 @@ def has_view_permission(self, user):
         return user.is_teacher or self.is_instructor(user) or self.is_registered(user)
 
     def has_edit_permission(self, user):
-        return user.is_teacher or self.is_instructor(user)
+        course_reg = get_course_registration(user, self)
+        return course_reg == INSTRUCTOR
```
keyvankhademi, line 134: shouldn't this be course_reg.registration_type == INSTRUCTOR?

keyvankhademi (changes requested):

keyvankhademi (approved):

PR #284, opened 2022-01-24
Add on to API for SubmissionAnalytics

sirusw:
- Rest api for viewing all submissionAnalytics or retrieve single model by submission id (based on Keyvan's work)
- Create SubmissionAnalytics upon accessing, if exists in db returns it, if not calculate dynamically and store in db
- add time_spent attribute to SubmissionAnalytics


analytics/migrations/0001_initial.py
```diff
@@ -1,6 +1,7 @@
-# Generated by Django 3.0.14 on 2022-01-23 19:55
```
keyvankhademi, line 1: Don't change the existing migration files.

analytics/migrations/0002_submissionanalytics_time_created.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 3.0.14 on 2022-01-24 03:12
```
keyvankhademi, line 1: You should only need to add 1 migration. All the changes should be here.

analytics/migrations/0003_auto_20220123_2309.py
```diff
@@ -0,0 +1,36 @@
+# Generated by Django 3.0.14 on 2022-01-24 07:09
```
keyvankhademi, line 1: only 1 migration file.

analytics/models.py
```diff
@@ -1,4 +1,4 @@
-from django.db import models
+rom django.db import models
```
keyvankhademi, line 1: I believe this is a mistake.

analytics/models/java.py
```diff
@@ -0,0 +1,25 @@
...
+from django.db import models
+from course.fields import JSONField
+
+
+class JavaSubmissionAnalytics(SubmissionAnalytics):
+    lines = models.IntegerField(default=0)
+    blank_lines = models.IntegerField(default=0)
+    comment_lines = models.IntegerField(default=0)
+    import_lines = models.IntegerField(default=0)
+    cc = models.IntegerField(default=0)
```
keyvankhademi, line 11: We need to have better names. All the names should be meaningful. `cc` is not a good name, if it's an abbreviation please write it in expanded form.

analytics/models/java.py
```diff
@@ -0,0 +1,25 @@
...
+    method = models.IntegerField(default=0)
```
keyvankhademi, line 12: better name

analytics/models/java.py
```diff
@@ -0,0 +1,25 @@
...
+    operator = models.IntegerField(default=0)
```
keyvankhademi, line 13: better name

analytics/models/java.py
```diff
@@ -0,0 +1,25 @@
...
+    operand = models.IntegerField(default=0)
```
keyvankhademi, line 14: better name

analytics/models/java.py
```diff
@@ -0,0 +1,25 @@
...
+    unique_operator = models.IntegerField(default=0)
+    unique_operand = models.IntegerField(default=0)
+    vocab = models.IntegerField(default=0)
```
keyvankhademi, line 17: better name. Most of these names are only meaningful to you. Try to name them in a way that anyone can understand what they are. If some cases if you can't come up with a name at least provide a description.

analytics/models/java.py
```diff
@@ -0,0 +1,25 @@
...
+    size = models.IntegerField(default=0)
```
keyvankhademi, line 18: better name.

analytics/models/java.py
```diff
@@ -0,0 +1,25 @@
...
+    vol = models.DecimalField(max_digits=8, decimal_places=4, default=0)
```
keyvankhademi, line 19: better name

analytics/models/java.py
```diff
@@ -0,0 +1,25 @@
...
+    difficulty = models.DecimalField(max_digits=8, decimal_places=4, default=0)
+    effort = models.DecimalField(max_digits=8, decimal_places=4, default=0)
+    error = models.DecimalField(max_digits=8, decimal_places=4, default=0)
```
keyvankhademi, line 22: I have no idea what these terms mean.

keyvankhademi: I think a description needs to be provided here.

analytics/services/submission_analytics.py
```diff
@@ -1,22 +1,229 @@
+from accounts.models import MyUser
```
keyvankhademi, line 1: This file is too big. You have to split it into multiple files.

analytics/services/submission_analytics.py
```diff
@@ -1,22 +1,229 @@
...
+                                                           operator=sub_analytics_dict.operator,
+                                                           operand=sub_analytics_dict.operand,
+                                                           unique_operator=sub_analytics_dict.unique_operator,
+                                                           unique_operand=sub_analytics_dict.unique_operand,
+                                                           vocab=sub_analytics_dict.vocab,
+                                                           size=sub_analytics_dict.size, vol=sub_analytics_dict.vol,
+                                                           difficulty=sub_analytics_dict.difficulty,
+                                                           effort=sub_analytics_dict.effort,
+                                                           error=sub_analytics_dict.error,
+                                                           test_time=sub_analytics_dict.test_time)
```
keyvankhademi, line 84: Find a way to avoid these huge blocks. Maybe a constructor for the model would be useful.

analytics/tests.py
```diff
@@ -1 +1,224 @@
...
+from course.utils.utils import create_mcq_submission, create_multiple_choice_question
+from test.base import BaseTestCase
+from .models import MCQSubmissionAnalytics, JavaSubmissionAnalytics, ParsonsSubmissionAnalytics
+from .models.models import SubmissionAnalytics
+from .services.submission_analytics import get_all_submission_analytics
+from .utils import init_analytics
+
+
+class SubmissionAnalyticsTestCase(BaseTestCase):
+    TEST_CODE1 = """public class Test
```
keyvankhademi, line 16: These codes should be in separate files.

api/urls.py
```diff
@@ -43,7 +44,7 @@
 router.register(r'course-admin', CourseAdminViewSet, basename='admin-course')
 router.register(r'uqj-update', UpdateUQJViewSet, basename='uqj-update')
 router.register(r'question-report', QuestionReportViewSet, basename='question-report')
-router.register(r'analytics', AnalyticsViewSet, basename='analytics')
+router.register(r'submission-analytics', AnalyticsViewSet, basename='submission-analytics')
```
keyvankhademi, line 47: don't change this. the paths for the apis should be like this. `/analytics/submission` or `/analytics/question`

api/views/question.py
```diff
@@ -79,7 +79,7 @@ def get_favorite_count(self, request, pk=None):
         return Response(uqj_count)
 
     # TODO: Fix permissions for this
-    @action(detail=True, methods=['post'], url_path='opened-question')
+    @action(detail=True, methods=['get'], url_path='opened-question')
```
keyvankhademi, line 82: dont' change the method.

course/migrations/0025_submission_time_spent.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.0.14 on 2022-01-13 02:28
+
```
keyvankhademi, line 2: remove the migrations.

course/migrations/0026_remove_submission_time_spent.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 3.0.14 on 2022-01-24 07:23
+
+from django.db import migrations
+
```
keyvankhademi, line 4: remove the migrations.

keyvankhademi (changes requested):

analytics/migrations/0002_submissionanalytics_time_created.py
sirusw: Done.

analytics/models.py
sirusw: Fixed

analytics/models/java.py
sirusw: Fixed.

api/urls.py
sirusw: Fixed.

api/views/question.py
sirusw: I tried with post but it actually doesn't work, it says the method is not allowed

analytics/services/submission_analytics.py
sirusw: I tried with a constructor but doesn't seem like this could be avoided, but I figured out that there are some redundancy in this file so I deleted a lot of code.

analytics/services/submission_analytics.py
sirusw: I removed a lot of redundant code, should be better now.

PR #300, opened 2022-04-26
Fix Permission Error

### LINKED ISSUE #302: Fix Permission Error

carson-ricca:
- This seems to fix the error, I think the permission class was left in after other changes were made. Need to ensure the UI process works fully first.
- Closes #302

### DIFF
--- a/api/views/course.py
+++ b/api/views/course.py
@@ -6,7 +6,7 @@ from rest_framework.generics import get_object_or_404
 from rest_framework.response import Response
 
 from api.serializers import CourseSerializer, CourseSerializerList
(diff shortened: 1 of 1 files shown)


keyvankhademi:
What are we gonna do about this PR? @carson-ricca

carson-ricca:
> What are we gonna do about this PR? @carson-ricca 

Good question, when I was testing it I didn't find anything that appeared to work incorrectly with this change. The change did fix the issue that was found. I just don't know enough about the intention of this permission to know if I'm missing any use cases for it but I think it's not needed anymore...

keyvankhademi:
@carson-ricca I don't think this is the problem. The permission seems to be right, a user who is not registered in a course should not be able to see it. So I'll close this for now. I'll look into the problem.

PR #309, opened 2022-07-04
Update Junit JAR to V1.2.0

Hedgemon4:


canvas-gamification.iml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
keyvankhademi, line 1: I think you accidentally committed this file. Please remove it if that's the case.

keyvankhademi (changes requested):

PR #320, opened 2022-10-13
Leaderboard for challenge

### LINKED ISSUE #319: Add leader_board property for each event

yuqi88:
closes #319

### DIFF
--- a/canvas/models/models.py
+++ b/canvas/models/models.py
@@ -309,6 +309,16 @@ class Event(models.Model):
         return score
 
     @property
+    def leader_board(self):
+        if self.type == "CHALLENGE":
+            return [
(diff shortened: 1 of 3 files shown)


canvas/models/models.py
```diff
@@ -308,6 +308,16 @@ def calculate_score(self, team):
 
         return score
 
+    @property
+    def leader_board(self):
```
keyvankhademi, line 312: Why do we need this? Can't you just get the challenge teams?

keyvankhademi (changes requested):
Please use the teams. There is no need to add a new property.

keyvankhademi (approved):

PR #324, opened 2022-11-17
Add member_names to Team model

yuqi88:
closes #323

PR #325, opened 2022-11-23
Add Course api - registered_students

yuqi88:
Please merge PR #324 first before merging this PR (this is a branch based on #324).

Add `registered_students`. 
Given course_id, return all students' info from courseRegistration (only return courseRegistration id & name)


canvas/models/models.py
```diff
@@ -157,7 +157,10 @@ def username(self):
 
     @property
     def name(self):
-        return self.user.get_full_name()
+        # registering in a course will force students to finish the profile?
```
keyvankhademi, line 160: no it doesn't.

api/views/course.py
```diff
@@ -126,3 +127,16 @@ def user_stats(self, request, pk=None, category_pk=None):
                 break
 
         return Response({"success_rate": success_rate})
+
+    @action(detail=False, methods=["get"], url_path="registered-students")
+    def registered_students(self, request):
```
keyvankhademi, line 132: this function name is misleading. it's returning course_registrations not students.
rename this to `course_registrations`.

api/services/course.py
```diff
@@ -0,0 +1,6 @@
+from canvas.models.models import CanvasCourse
+
+
+def get_registered_students(course: CanvasCourse):
```
keyvankhademi, line 4: You don't need a service for this. Just put it in the view.

api/views/course.py
```diff
@@ -126,3 +126,16 @@ def user_stats(self, request, pk=None, category_pk=None):
...
+        """
+        Given course id, return all students within a class
+        """
+        course_id = request.Get.get("course_id", None)
+        course = get_object_or_404(CanvasCourse, id=course_id)
+
+        course_regs = [{course_reg.id, course_reg.name} for course_reg in course.canvascourseregistration_set.all()]
+
+        serializer = self.get_serializer(course_regs)
+        return Response(serializer.data)
```
keyvankhademi, line 141: How are you not getting an error here?

keyvankhademi (approved):

PR #338, opened 2023-01-13
Add logging for teams and challenges

### LINKED ISSUE #337: Add logging for teams and challenges
- Log team creation
- Log joining team
- Log challenge completion of teams

yuqi88:
- Log team creation
- Log joining team
- Log challenge completion of teams

closes #337

### DIFF
--- a/api/views/submission.py
+++ b/api/views/submission.py
@@ -140,11 +140,10 @@ class SubmissionViewSet(viewsets.GenericViewSet):
                 ]
 
                 solved_event_question_id = list(set(solved_event_question_id))
-                event_question_ids = [ele['id'] for ele in event.question_set.values_list('id')]
+                event_question_ids = [ele["id"] for ele in event.question_set.values_list("id")]
 
                 if len(event_question_ids) is len(solved_event_question_id):
                     team_complete_challenge_action(question.event.id, team, request.user)
-
 
         return Response(
             self.get_serialized_data(submission),
--- a/api/views/submission.py
+++ b/api/views/submission.py
@@ -22,9 +22,10 @@ from api.serializers.multiple_choice_question import (
 from api.serializers.parsons_question import (
     ParsonsSubmissionHiddenDetailsSerializer,
 )
+from canvas.models.models import EVENT_TYPE_CHOICES
 from course.exceptions import SubmissionException
 from course.models.java import JavaQuestion, JavaSubmission
-from course.models.models import Submission, Question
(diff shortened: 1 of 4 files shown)


general/services/action.py
```diff
@@ -324,3 +324,42 @@ def create_goal_item_action(goal_item, user):
...
+def create_team_action(team, user):
+    Action.create_action(
+        actor=user,
+        description="User created a team.",
+        token_change=0,
+        status=ActionStatus.COMPLETE,
+        verb=ActionVerb.CREATED,
+        object_type=ActionObjectType.TEAM,
+        object_id=team.id,
+        data=team,
```
keyvankhademi, line 338: make sure the data is stored correctly. You can check it by going in the admin panel and see what is actually stored.

api/views/submission.py
```diff
@@ -121,6 +122,29 @@ def submit(self, request):
...
+                    solved_uqj.question.id
+                    for solved_uqj in solved_uqjs
+                    if solved_uqj.question.event.id is question.event.id
+                ]
+
+                solved_event_question_id = list(set(solved_event_question_id))
+                event_question_ids = [ele["id"] for ele in event.question_set.values_list("id")]
+
+                if len(event_question_ids) is len(solved_event_question_id):
+                    team_complete_challenge_action(question.event.id, team, request.user)
```
keyvankhademi, line 146: This is a significant overhead for each submission. Also, it throws an error if a question is not in an event.
I think we can infer this data from what we currently store in this way:
You get all the submissions for an event. Then you check if all the questions have a correct submission and if yes then you get the last submission (based on time) and that is exactly what you are doing here.

keyvankhademi (changes requested):

general/models/action.py
```diff
@@ -43,6 +44,7 @@ class ActionObjectType:
     QUESTION = "Question"
     SUBMISSION = "Submission"
     USER = "User"
+    TEAM = "Team"
```
keyvankhademi, line 47: These require a migration. Did you not face an error?

keyvankhademi (changes requested):

general/models/action.py
yuqi88: It didn't give me any errors.
I ran the migration command, but I still have this:

image

keyvankhademi: Have you created the migrations?
you have to run `python manage.py createmigrations` first and then `python manage.py migrate`.

keyvankhademi (approved):

PR #340, opened 2023-01-16
Add course and event leader board APIs

### LINKED ISSUE #339: Add course and event leader board APIs

yuqi88:
- remove leader_board property from the course model and make it an API 
- add event-leader_boards API in event
closes #339


api/views/event.py
```diff
@@ -119,3 +119,30 @@ def import_event(self, request):
             self.get_serializer(cloned_event).data,
             status=status.HTTP_201_CREATED,
         )
+
+    @action(detail=True, method=["get"], url_path="event-leader-boards")
+    def event_leader_boards(self, request, pk):
```
keyvankhademi, line 124: name should be `leader_board` when it's in the course it's the course leader board and when it's in an event it's the event leader board.

/api/event/1/leader-board
/api/course/1/leader-board

keyvankhademi (changes requested):

api/views/event.py
yuqi88: I'm trying to get all leaderboards for all events of a course.
That way we don't have to call the API every time a student switches to see another challenge. Is this a bad idea?

keyvankhademi: In that case, it should be in the course and the url will be `/api/course/1/event-leader-boards`.
As the leaderboard calculation is significant. it might still be better to request them 1 by 1 as we are not showing them at the same time.

keyvankhademi:
Let's not remove the leaderboard for now because it will break the f/e and prevents us from releasing it.
also, pull from the master I made some changes to not break the f/e

api/views/course.py
```diff
@@ -137,3 +137,20 @@ def course_registrations(self, request, pk):
...
+        Given course id, return the leader board for the course.
+        """
+        course = get_object_or_404(CanvasCourse, id=pk)
+        leader_board = [
+            {
+                "name": course_reg.user.get_full_name(),
+                "token": course_reg.total_tokens_received,
+            }
+            for course_reg in course.canvascourseregistration_set.all()
+            if course_reg.is_verified
```
keyvankhademi, line 153:
```suggestion
            for course_reg in self.canvascourseregistration_set.filter(status="VERIFIED").all()
```

keyvankhademi (approved):

PR #345, opened 2023-02-07
Changed the course URL to not be required by the course

Hedgemon4:


canvas/models/models.py
```diff
@@ -14,7 +14,7 @@
 
 class CanvasCourse(models.Model):
     name = models.CharField(max_length=500)
-    url = models.URLField()
+    url = models.URLField(null=True, blank=True)
```
keyvankhademi, line 17: This requires a migration.

keyvankhademi (changes requested):

keyvankhademi (approved):

PR #351, opened 2023-02-14
Recalcuate leaderboard tokens

### LINKED ISSUE #350: course leaderboard needs to count practice tokens

yuqi88:
close #350

### DIFF
--- a/canvas/models/models.py
+++ b/canvas/models/models.py
@@ -135,7 +135,12 @@ class CanvasCourseRegistration(models.Model):
 
         from course.models.models import UserQuestionJunction
 
-        practiced_uqjs = UserQuestionJunction.objects.filter(user=self.user, is_solved=True, question__event=None)
+        practiced_uqjs = UserQuestionJunction.objects.filter(
+            user=self.user,
+            question__course=None,
+            question__is_verified=True,
+            question__status="CREATED",
+        )
         for uqj in practiced_uqjs:
             tokens += uqj.tokens_received
         # print('token: ', tokens) # token: xx is printed multiple times, and eventually ends at 0. why?
--- a/canvas/models/models.py
+++ b/canvas/models/models.py
@@ -132,6 +132,14 @@ class CanvasCourseRegistration(models.Model):
                 tokens += 0
             else:
                 tokens += event.tokens_received(team)
+
+        from course.models.models import UserQuestionJunction
+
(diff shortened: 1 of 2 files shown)


canvas/models/models.py
```diff
@@ -132,6 +132,14 @@ def total_tokens_received(self):
                 tokens += 0
             else:
                 tokens += event.tokens_received(team)
+
+        from course.models.models import UserQuestionJunction
+
+        practiced_uqjs = UserQuestionJunction.objects.filter(user=self.user, is_solved=True, question__event=None)
```
keyvankhademi, line 138: The filters are incorrect. You dont need is solved as there might be partially solved too. You need question course = None and question status to be created and is verified to be true. Basically the same filters that i told you in one of the other PRS.

keyvankhademi: Also try to avoid local imports. You can probably move the to the service folder of the course app.

canvas/models/models.py
```diff
@@ -132,6 +132,14 @@ def total_tokens_received(self):
...
+        for uqj in practiced_uqjs:
+            tokens += uqj.tokens_received
+        # print('token: ', tokens) # token: xx is printed multiple times, and eventually ends at 0. why?
```
keyvankhademi, line 141: Check why this method is being called multiple times. Are they all for the same user? What are the uqjs when the token is 0?

keyvankhademi (changes requested):

canvas/models/models.py
```diff
@@ -132,6 +132,19 @@ def total_tokens_received(self):
                 tokens += 0
             else:
                 tokens += event.tokens_received(team)
+
+        from course.models.models import UserQuestionJunction
+        from course.models.models import Question
+
+        practiced_uqjs = UserQuestionJunction.objects.filter(
```
keyvankhademi, line 139: You can use `self.user.question_junctions`

canvas/models/models.py
```diff
@@ -132,6 +132,19 @@ def total_tokens_received(self):
...
+            user=self.user,
+            question__course=None,
```
keyvankhademi, line 141: You still need question_event=None

keyvankhademi (changes requested):

keyvankhademi (approved):

PR #363, opened 2023-03-21
Fix duplicate event question name when adding

### LINKED ISSUE #352: Fix duplicate event question names
If we have multiple categories each category will have numbers 1..n. causing a bunch of duplicates.

yuqi88:
closes #352


canvas/services/event.py
```diff
@@ -73,6 +74,14 @@ def set_featured(event):
 
 
 def add_question_set(event, category_id, difficulty, number_of_questions):
+    def begins_with_number(string):
+        # return any(char.isdigit() for char in string)
+        return string[0].isdigit()
```
keyvankhademi, line 79: Let's not do this. If you are searching for numbers in the title it doesn't have to be at the beginning.

keyvankhademi: you can remove this function.

canvas/services/event.py
```diff
@@ -73,6 +74,14 @@ def set_featured(event):
...
+
+    def extract_1st_number(string):
+        # return re.search('[0-9]+', string).group()
+        return re.findall(r"\d+", string)[0]
```
keyvankhademi, line 83: make this functions error proof so if there is not digit return something like -1 or 0

canvas/services/event.py
```diff
@@ -82,6 +91,20 @@ def add_question_set(event, category_id, difficulty, number_of_questions):
         difficulty=difficulty,
     )[:number_of_questions]
 
-    index = 0  # in case items is empty and you need it after the loop
-    for index, question in enumerate(questions, start=1):
-        question.copy_to_event(event, str(index))
+    used_titles = [question["title"] for question in event.question_set.values("title")]
+    used_titles = [extract_1st_number(title) for title in used_titles if begins_with_number(title)]
```
keyvankhademi, line 95: here you can just extract the number

canvas/services/event.py
```diff
@@ -82,6 +91,20 @@ def add_question_set(event, category_id, difficulty, number_of_questions):
...
+    used_titles.sort()
+    print(used_titles)
```
keyvankhademi, line 97: no need to sort or print

canvas/services/event.py
```diff
@@ -82,6 +91,20 @@ def add_question_set(event, category_id, difficulty, number_of_questions):
...
+
+    if len(used_titles) == 0:
+        idx = 0  # in case items is empty, and you need it after the loop
+        for idx, q in enumerate(questions, start=1):
+            q.copy_to_event(event, str(idx))
+        return
```
keyvankhademi, line 103: i don't think this is neccessary

canvas/services/event.py
```diff
@@ -82,6 +91,20 @@ def add_question_set(event, category_id, difficulty, number_of_questions):
...
+
+    title = 1
+    for idx, q in enumerate(questions):
+        while used_titles.count(str(title)) != 0:
```
keyvankhademi, line 107:
```suggestion
        while str(title) in used_titles:
```

keyvankhademi (changes requested):

canvas/services/event.py
```diff
@@ -82,6 +87,12 @@ def add_question_set(event, category_id, difficulty, number_of_questions):
...
     )[:number_of_questions]
 
-    index = 0  # in case items is empty and you need it after the loop
-    for index, question in enumerate(questions, start=1):
-        question.copy_to_event(event, str(index))
+    used_titles = [question["title"] for question in event.question_set.values("title")]
+    used_titles = [extract_1st_number(title) for title in used_titles]
+
+    title = 1
+    for idx, q in enumerate(questions):
```
keyvankhademi, line 94: are u using idx?
```suggestion
    for q in questions:
```

keyvankhademi (changes requested):

keyvankhademi (approved):

PR #366, opened 2023-04-10
My stats challenge support

yuqi88:


keyvankhademi:
@yuqi88 What's going on with this? Please either complete this or close it.

yuqi88:
@keyvankhademi  I have updated the top comment and added a screenshot to describe the PR.

api/services/stats.py
```diff
@@ -69,8 +70,11 @@ def get_question_stats(user):
 
 
 def get_challenge_stats(user):
+    events = Event.objects.filter(type="CHALLENGE")
+    challenges_solved = sum(event.has_solved_event(user) for event in events)
```
Opeyem1a, line 74: The logic here seems fine but as a sanity check, what is `challenges_solved` supposed to mean?

`Event.has_solved_event()` only checks if any of the questions within that event have been solved by the user - the variable name `challenges_solved` implied to me that it means the number of challenges they have solved every question from, but the value you're calculating doesn't seem to be that

Opeyem1a (changes requested):

api/services/stats.py
Opeyem1a: To clarify here, I'm basically saying I think there is a logic error in what `Event.has_solved_event()` does as a function

canvas/models/models.py
```diff
@@ -274,7 +274,7 @@ def has_solved_event(self, user):
 
         uqjs = UserQuestionJunction.objects.filter(user_id=user.id, is_solved=True, question__in=event_questions)
 
-        if uqjs.exists():
+        if uqjs.count() == event_questions.count():
             return True
         return False
```
yuqi88, line 279: I'm not sure if there is a better way to do this. 
```
if xxx:
   return True
return False
```
feels not as clean. Opening to advice!

Opeyem1a: I think this code is fine tbh, if I was to nitpick I suppose you could just `return uqjs.count() == event_questions.count()` instead of the conditional and it would logically be the same though

Opeyem1a (approved):

PR #378, opened 2023-07-17
make mcq stats labels non varied text

SamamthaH:
image


SamamthaH:
In this case the only thing in the correct answer and distractor text box was the variable, however it could also look like "the number: {{num}} is even"

canvas/services/event.py
```diff
@@ -44,8 +45,9 @@ def get_question_stats(question):
 
     answers = {}
     if isinstance(question, MultipleChoiceQuestion):
+        choices = json.loads(question.choices) if type(question.choices) == str else question.choices
```
keyvankhademi, line 48: In what situation do you get `question.choices` to be str?

SamamthaH: I am unsure, I took this line from course/models/models.py get_rendered_choices (line 323 specifically) which the original code called instead. I presume it is in there for a reason so I didn't want to mess with it.

keyvankhademi (approved):

keyvankhademi:
@SamamthaH Please fix the lint

SamamthaH:
I have no idea what the gunicorn file is but apparently that was the thing that needed reformatting. Feels weird to have it tacked on to this pr though.

PR #380, opened 2023-07-25
Add question details to gradebook

SamamthaH:


api/views/course.py
```diff
@@ -184,20 +184,33 @@ def course_grade_book(self, request, pk):
         for event in events.all():
             teams.extend(event.team_set.all())
 
+        from course.models.models import UserQuestionJunction, Submission
```
keyvankhademi, line 187: The imports should be at the top. It that causes circular import we probably have done something wrong.

api/views/course.py
```diff
@@ -184,20 +184,33 @@ def course_grade_book(self, request, pk):
...
-                        "name": course_reg.name,
-                        "event_name": team.event.name,
-                        "legal_first_name": consent.legal_first_name if consent else "",
-                        "legal_last_name": consent.legal_last_name if consent else "",
-                        "student_number": consent.student_number if consent else "",
-                    }
-                )
+                question_ids = [question.id for question in team.event.question_set.all().order_by('title')]
+                for question_id in question_ids:
+                    uqj = UserQuestionJunction.objects.get(id=question_id)
```
keyvankhademi, line 195:
```suggestion
                    uqjs = question.question_junctions
```

keyvankhademi: or it might be something similar to that I'm not sure about the exact term

api/views/course.py
```diff
@@ -184,20 +184,33 @@ def course_grade_book(self, request, pk):
...
+                    submissions = Submission.objects.filter(uqj__question__id=question_id, uqj__user=course_reg.user)
```
keyvankhademi, line 196:
```suggestion
                    submissions = uqj.submissions
```

api/views/course.py
SamamthaH: The closest I could find was question.user_junctions which as far as I can tell only have the fields: 
id, user_id, question_id, random_seed, last_viewed, opened_tutorial, tokens_received, solved_at, is_solved, is_partially_solved, is_favorite
I can't find something that will give me access to the submissions. Is there a particular place I should be looking that I haven't yet?

api/permissions.py
```diff
@@ -51,7 +51,10 @@ def has_object_permission(self, request, view, obj):
 class GradeBookPermission(permissions.IsAuthenticated):
     def has_object_permission(self, request, view, obj):
         if request.method == "GET":
-            return obj.has_edit_permission(request.user)
+            if request.user.is_student:
```
Opeyem1a, line 54: I get this but it scares me in a future way - like this permission implies that just being a student that is registered should grant access to APIs with this permission

I know that in the endpoint itself, we only show students their own data, but if someone tried to reuse this permission based on its name, we have have some issues

Overall, maybe it's fine but definitely toss a # TODO comment on this and/or a ticket

(Bigger solution I would suggest is to have separate endpoints for the "see all gradebook data" vs "see MY gradebook data" and have different permissions for each, but in the interest of time, don't sweat this too too much)

Opeyem1a (commented):
it's a lot of nested for loops so a bit hard to follow but I think the logic is correct (?) - other than my other comments it seems good

Opeyem1a (approved):

api/permissions.py
keyvankhademi: You need to disallow students from the list API. The list API doesn't consider the obj permissions

api/views/course.py
```diff
@@ -182,12 +182,33 @@ def course_grade_book(self, request, pk):
...
+                )
+            else:
+                course_registrations = team.course_registrations.filter(registration_type="STUDENT", status="VERIFIED")
+            for course_reg in course_registrations:
+                question_details = []
                 consent = course_reg.user.consents.last()
+                questions = team.event.question_set.all().order_by("title")
+                for question in questions:
+                    submissions = question.user_junctions.get(user=course_reg.user).submissions.all()
+                    best_attempt_grade = max([submission.grade for submission in submissions]) if submissions else 0
```
keyvankhademi, line 204: If the logic of best attempt changes in the future this part will break and we will have no idea we have to fix it here until someone report the issue. It's better to use uqj's here too.

PR #381, opened 2023-08-11
Add most recent submissions to event stats

SamamthaH:


canvas/services/event.py
```diff
@@ -58,9 +78,10 @@ def get_question_stats(question):
         "has_variables": len(question.variables) > 0,
         "answers": answers,
         "error_messages": _get_error_messages(submissions),
-        "submissions": _get_submission_status(submissions),
+        "submissions": _get_submission_status(recent_submissions, student_course_registrations),
         "status_messages": _get_status_messages(submissions),
         "total_submissions": submissions.count(),
+        "student_course_registrations": student_course_registrations,
```
Opeyem1a, line 84: the naming comment applies here too

canvas/services/event.py
```diff
@@ -31,11 +33,12 @@ def _get_error_messages(submissions):
     return error_messages
 
 
-def _get_submission_status(submissions):
+def _get_submission_status(submissions, students):
```
Opeyem1a, line 36: the variable name `students` count probably be named better too

`students` kinda implies a list of Student, when you actually mean something more like `num_students`

Even better if you do `def _get_submission_status(submissions, num_students: int):`

Opeyem1a (changes requested):

Opeyem1a:
Also the lint is failing

Opeyem1a (approved):

canvas/services/event.py
```diff
@@ -32,11 +34,12 @@ def _get_error_messages(submissions):
     return error_messages
 
 
-def _get_submission_status(submissions):
+def _get_submission_status(submissions, num_students):
     return {
         "Correct": submissions.filter(is_correct=True).count(),
         "Partially Correct": submissions.filter(is_correct=False, is_partially_correct=True).count(),
         "Incorrect": submissions.filter(is_correct=False, is_partially_correct=False).count(),
+        "Not Attempted": num_students - submissions.count(),
```
keyvankhademi, line 42: This doesn't seem right. You are subtracting `number of submissions` from `number of students`. You can't subtract 2 numbers with different units. The only circumstance in which this makes sense is that the submissions are guaranteed to be the last submission of each student. Even if that's the case it's not obvious that this is a hidden assumption of this function.

canvas/services/event.py
```diff
@@ -52,6 +55,20 @@ def get_question_stats(question):
...
 
+    most_recent_submission_time_for_submission_author = (
+        submissions.filter(uqj__user=OuterRef("uqj__user")).order_by("-submission_time").values("submission_time")[:1]
+    )
+
+    recent_submissions = (
+        submissions.filter(uqj__user__role="Student")
+        .annotate(recent_timestamp=Subquery(most_recent_submission_time_for_submission_author))
+        .filter(submission_time=F("recent_timestamp"))
+    )
```
keyvankhademi, line 66: It is not guaranteed that the last submission is actually the one we consider the final submission. In most cases, we get the maximum of their grades. So if you submit a wrong code after a correct one, we consider the question to be solved correctly.

canvas/services/event.py
```diff
@@ -52,6 +55,20 @@ def get_question_stats(question):
...
+
+    num_students_registered_in_course = question.course.verified_course_registration.filter(
+        registration_type="STUDENT"
+    ).count()
```
keyvankhademi, line 70: Although we are only showing this in events, the functionality works for all the questions. Even the ones that are not in a course. This code will break for the questions that are not in a course.

SamamthaH:
_get_submission_status was only used for this api, so since I couldn't use it with this system of getting student progress i got rid of it as it had no other use

canvas/services/event.py
```diff
@@ -52,6 +44,29 @@ def get_question_stats(question):
...
+    for partial in partially_correct_submissions:
+        if not users.__contains__(partial["uqj__user_id"]):
+            progress_submission_statuses[1] += 1
+            users.append(partial["uqj__user_id"])
+
+    incorrect_submissions = student_submissions.filter(is_partially_correct=False, is_correct=False).values(
+        "uqj__user_id"
+    )
+    for incorrect in incorrect_submissions:
+        if not users.__contains__(incorrect["uqj__user_id"]):
```
Opeyem1a, line 66: Google this a tad but I think the `in` operator is what `__contains__` means.
Like https://www.online-python.com/d8HlpBGn2K

The methods with `__something__` structure are called "dunders" and they typically correlate with a keyword/symbol (i.e. `__getattribute__` corresponds to the `.` operator)

tldr; I think `in` should be used, not `__contains__`

canvas/services/event.py
```diff
@@ -52,6 +44,29 @@ def get_question_stats(question):
                 answers[answer] = 0
             answers[answer] += 1
 
+    users = []
+    progress_submission_statuses = [0, 0, 0]
```
Opeyem1a, line 48: This variable actually stores counts of students that correspond to various submission statuses, this variable name feels iffy based on this description.

Second ick is that it's a bit messy to understand which index corresponds to each status (but reading the code makes it fairly clear tbh) - still I'd probably have used a dictionary with keys that describe what the value is.
i.e. (kinda weak naming on my part here, adjust as makes sense)
```py
best_submission_counts = {
    "num_correct": 0,
    "num_partially_correct": 0,
    "num_incorrect": 0,
}
```

Especially since when we return this value we are returning a dictionary that names each of them anyway

canvas/services/event.py
```diff
@@ -52,6 +44,29 @@ def get_question_stats(question):
...
+    student_submissions = submissions.filter(uqj__user__role="Student")
+    correct_submissions = student_submissions.filter(is_correct=True).values("uqj__user_id")
+    for correct in correct_submissions:
+        if not users.__contains__(correct["uqj__user_id"]):
+            progress_submission_statuses[0] += 1
+            users.append(correct["uqj__user_id"])
```
Opeyem1a, line 54: nit;

What you did works and makes sense but I do think there is a cool way that uses python `set()` to do this better.

If you're curious, look into the `difference()` method of python sets - https://www.geeksforgeeks.org/python-set-difference/ and we can chat about it if what I'm saying sounds like gibberish

Haven't thought about it for too long but I have a hunch this approach saves us some big O complexity and might even save us a database call

canvas/services/event.py
```diff
@@ -52,6 +44,29 @@ def get_question_stats(question):
                 answers[answer] = 0
             answers[answer] += 1
 
+    users = []
+    progress_submission_statuses = [0, 0, 0]
+    student_submissions = submissions.filter(uqj__user__role="Student")
+    correct_submissions = student_submissions.filter(is_correct=True).values("uqj__user_id")
```
Opeyem1a, line 50: `.values_list("uqj__user_id", flat=True)` returns just the ids in an array, instead of a dictionary with the `"uqj__user_id"` key, would save you from have to do things like `correct["uqj__user_id"]`

Reference: https://stackoverflow.com/questions/37205793/django-values-list-vs-values

canvas/services/event.py
```diff
@@ -52,6 +44,29 @@ def get_question_stats(question):
...
+    for correct in correct_submissions:
+        if not users.__contains__(correct["uqj__user_id"]):
+            progress_submission_statuses[0] += 1
+            users.append(correct["uqj__user_id"])
+
+    partially_correct_submissions = student_submissions.filter(is_partially_correct=True).values("uqj__user_id")
```
Opeyem1a, line 56: nit;

For speed, we could adjust the filter so it's already ignoring the submissions from students with ids already in the `users` list.

Something like 
```py
student_submissions.filter(is_partially_correct=True).exclude(uqj__user_id__in=users).values("uqj__user_id")
```
_Note the `.exclude` and `__in`

Opeyem1a (changes requested):

Opeyem1a (approved):
Looks good to me 👌

canvas/services/event.py
```diff
@@ -52,6 +44,27 @@ def get_question_stats(question):
...
+        elif submission["is_partially_correct"]:
+            partially_correct_submissions_users.add(submission["uqj__user_id"])
+        else:
+            incorrect_submissions_users.add(submission["uqj__user_id"])
+
+    partially_correct_submissions_users = partially_correct_submissions_users.difference(correct_submissions_users)
+    incorrect_submissions_users = incorrect_submissions_users.difference(
+        correct_submissions_users, partially_correct_submissions_users
+    )
+
```
keyvankhademi, line 67: I still feel like this logic has flaws. Don't you include 2 entities for 1 user that made 2 submissions?

keyvankhademi: I think you should use uqj's here. get all the uqjs for the course and then you can filter them based on solved, partially solved, etc.

Opeyem1a (approved):

keyvankhademi (approved):

PR #385, opened 2023-09-21
Add Nickname to MyUser

Hedgemon4:


Hedgemon4:
See #727 for this task

PR #386, opened 2023-09-22
Add nickname to student object

ketphan02:
### Issue
We want to have a nickname for student to display instead of their legal name
This PR only work the back-end part of this change

### Changes
+ Added nickname field in model
+ Added nickname field to API output for courses and view profile
+ Added migration with generated nickname

### Example response
```
{
    "id": 3,
    "first_name": "123",
    "last_name": "321",
    "email": "123@abc.com",
    "nickname": "123"
}
```
When returning user profile of this user
image


accounts/models.py
```diff
@@ -26,6 +26,7 @@ def is_teacher(self):
 class MyUser(AbstractUser):
     role = models.CharField(max_length=100, choices=USER_ROLE_CHOICES, default=STUDENT)
     email = models.EmailField("email address", blank=True, unique=True)
+    nickname = models.CharField("nickname", max_length=30, blank=True)
```
Hedgemon4, line 29: You added the nickname field twice here.

Hedgemon4 (changes requested):

accounts/models.py
```diff
@@ -43,6 +44,10 @@ def is_student(self):
     def has_name(self):
         return self.first_name is not None and self.first_name != ""
 
+    @property
+    def has_nickname(self):
+        return self.nickname is not None and self.nickname != ""
+
```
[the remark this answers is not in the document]
ketphan02: This would be handy for later when returning information of people have not set their nicknames

accounts/models.py
ketphan02: The other one is UserConsent, which I should not add

ketphan02:
TODO: Discuss migration issue for nickname

accounts/models.py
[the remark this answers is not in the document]
Opeyem1a: Was about to comment the same thing as Seth. Currently it's enforced on the DB level that nickname cannot be null, right?

canvas/models/models.py
```diff
@@ -175,9 +176,9 @@ def username(self):
 
     @property
     def name(self):
-        if self.user.has_name:
-            return self.user.get_full_name()
-        return "Anonymous Student"
+        if self.user.has_nickname:
```
Opeyem1a, line 179: but it can't be null right?

canvas/models/models.py
```diff
@@ -175,9 +176,9 @@ def username(self):
...
+            return self.user.nickname
+        return generate_default_name()
```
Opeyem1a, line 181: If you're doing this because the nickname could be an empty string, thing returning a freshly generated nickname would be a bad idea, no?

It would be different every time and it's not saving this newly generated default nickname to the database

accounts/migrations/0016_myuser_nickname.py
```diff
@@ -0,0 +1,32 @@
+from django.db import migrations, models
+from accounts.utils.generate_default_name import generate_default_name
+
+
+def forwards_func(apps, schema_editor):
+    MyUser = apps.get_model("accounts", "myuser")
+    my_users = MyUser.objects.select_for_update()
+    for my_user in my_users:
+        my_user.nickname = generate_default_name()
```
Opeyem1a, line 9: they need to be unique, and that should be enforced in this case (i.e. keep generating until it doesn't clash with an existing nickname in the course)

accounts/models.py
```diff
@@ -43,6 +44,10 @@ def is_student(self):
     def has_name(self):
         return self.first_name is not None and self.first_name != ""
 
+    @property
+    def has_nickname(self):
+        return self.nickname is not None and self.nickname != ""
```
Opeyem1a, line 49: I feel like a better solution is to just not let them set an empty string nickname instead of guarding against this case in so many places

Opeyem1a (changes requested):

canvas/models/models.py
ketphan02: Technically, but this is for the very worst case

Opeyem1a: But if we have the integrity constraint on the DB, then it's not the worst case, it's an impossible case, no?

api/serializers/register.py
```diff
@@ -18,6 +18,18 @@ class UserRegistrationSerializer(serializers.ModelSerializer):
             )
         ],
     )
+    legal_first_name = serializers.CharField(
```
Hedgemon4, line 21: Why is this legal first name here, and not just first name?

api/serializers/register.py
```diff
@@ -18,6 +18,18 @@ class UserRegistrationSerializer(serializers.ModelSerializer):
...
+        required=True,
+        error_messages=ERROR_MESSAGES.FIRSTNAME.ERROR_MESSAGES,
+    )
+    legal_last_name = serializers.CharField(
```
Hedgemon4, line 25: Same here

Hedgemon4 (changes requested):

Opeyem1a (approved):
lgtm

Hedgemon4 (approved):

canvas/models/models.py
```diff
@@ -175,9 +175,10 @@ def username(self):
 
     @property
     def name(self):
-        if self.user.has_name:
-            return self.user.get_full_name()
-        return "Anonymous Student"
+        if not self.user.nickname:
+            raise ValueError(f"User {self.user.id} does not have a nickname")
```
keyvankhademi, line 179: This is not a good practice, we should not raise error. If the nickname field is `NOT NULL` in the database it will always have value.

api/serializers/register.py
```diff
@@ -37,7 +49,7 @@ class UserRegistrationSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = MyUser
-        fields = ("email", "password", "password2", "recaptcha_key")
+        fields = ("email", "first_name", "last_name", "nickname", "password", "password2", "recaptcha_key")
 
     def create(self, validated_data):
         user = MyUser.objects.create_user(username=validated_data["email"], email=validated_data["email"])
```
keyvankhademi, line 55: You should add the fields `first_name`, ... here to save it

keyvankhademi (approved):

keyvankhademi (approved):

PR #403, opened 2025-05-20
Gradebook export

SamamthaH:
added an export api call for the instructor's gradebook and updated leaderboard return to send a flag which indicates the potential token discrepancy between the Gradebook and the Leaderboard.

### DIFF
--- a/api/views/course.py
+++ b/api/views/course.py
@@ -195,7 +195,7 @@ class CourseViewSet(viewsets.ModelViewSet):
                     "question_details": [
                         {
                             "title": uqj.question.title,
-                            "question_grade": uqj.tokens_received,
+                            "question_grade": uqj.formatted_current_tokens_received,
                             "attempts": uqj.submissions.count(),
                             "max_attempts": uqj.question.max_submission_allowed,
                         }
@@ -223,7 +223,7 @@ class CourseViewSet(viewsets.ModelViewSet):
                         "question_details": [
                             {
                                 "title": uqj.question.title,
-                                "question_grade": uqj.tokens_received,
+                                "question_grade": uqj.formatted_current_tokens_received,
                                 "attempts": uqj.submissions.count(),
                                 "max_attempts": uqj.question.max_submission_allowed,
                             }
--- a/api/views/course.py
+++ b/api/views/course.py
@@ -18,6 +18,7 @@ from api.serializers.course import CourseCreateSerializer
 from api.serializers.eventSet import EventSetSerializer
 from canvas.models.models import CanvasCourse, Event
 from canvas.services.course import register_instructor
+from canvas.services.gradebook import get_student_gradebook, get_course_gradebook
 from canvas.utils.utils import get_course_registration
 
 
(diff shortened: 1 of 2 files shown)


api/views/course.py
```diff
@@ -206,7 +210,7 @@ def my_grades(self, request, pk):
         return Response(results)
 
     @action(detail=True, methods=["get"], url_path="grade-book", permission_classes=[GradeBookPermission])
-    def course_grade_book(self, request, pk):
+    def course_grade_book(self, request, pk, response=True):
```
keyvankhademi: you should not use a view and a function that returns data. You should extract the getting grade book logic to another file (probably in utils or services folder of the canvas app). And use that function in both views.

api/views/course.py
```diff
@@ -223,13 +227,98 @@ def course_grade_book(self, request, pk):
...
-        return Response(results)
+    @action(detail=True, methods=["get"], url_path="export-grade-book", permission_classes=[GradeBookPermission])
+    def export_grade_book(self, request, pk):
+        """
+        Optional Parameters:\n
+        ?event_name: string => filters retrieved grades by the event\n
+        ?student_name: string => filters retrieved grades by students whose name is a full or partial match\n
+        ?details: boolean => if true, adds question level breakdown to each overall event grade
+        """
+        event_name = request.GET.get("event_name", "")
```
keyvankhademi:
```suggestion
        event_name = request.GET.get("event_name", None)
```

api/views/course.py
```diff
@@ -223,13 +227,98 @@ def course_grade_book(self, request, pk):
...
+
+        response["Content-Disposition"] = (
+            f'attachment; filename="{self.get_object().name} '
+            f'{"" if student_name == "" else student_name + " students "}'
+            f'{"course" if event_name == "" else event_name} '
+            f'gradebook{" detailed" if details == "true" else ""}.csv"'
+        )
+        csv_data = self.course_grade_book("request", "pk", False)
+
+        if event_name != "" and event_name is not None:
```
keyvankhademi:
```suggestion
        if event_name is not None:
```

api/views/course.py
```diff
@@ -195,7 +199,7 @@ def my_grades(self, request, pk):
                     "question_details": [
                         {
                             "title": uqj.question.title,
-                            "question_grade": uqj.tokens_received,
+                            "question_grade": uqj.formatted_current_tokens_received,
```
keyvankhademi: i dont think you should use the formatted grade here since this is an API and can/will be used programmatically.
You have 2 options
1. add a new field `question_grade_formatted`.
2. Format the grade when writing to csv.

api/views/course.py
SamamthaH: I extracted it to the canvas services folder since use case seemed similar to the stats functions in the event file located there. I am curious if we have a specific system for where we are placing things between the different folders (api vs canvas vs course) and services vs utils

keyvankhademi (approved):

